Issue #4, opened 2021-11-26
Re-using the port after stopping seems to cause trouble


lawik:
I get enough warning to push any errors out of my scroll-back if I try to start a new pipeline on that same port after I've shut down the previous one.

Steps:
- Start a pipeline
- Stream media to it
- Stop it and shut it down
- Attempt to start it again

Not sure about the details but I believe this was seen frequently during that livestream I did.

PR #7, opened 2021-12-01
Bugfixes: Source refusing to terminate and honoring timestamps

daniel-jodlos:
Closing #4


lib/membrane_rtmp_plugin/rtmp/source.ex
```diff
@@ -58,10 +66,40 @@ defmodule Membrane.RTMP.Source do
...
+        {:empty, queue} -> {[], queue}
+        {{:value, buffer}, queue} -> {[buffer: {type, buffer}, redemand: type], queue}
+      end
+
+    state = put_in(state, [:buffers, type], queue)
+    {actions, state}
+  end
+
+  @impl true
+  def handle_other({:result, result}, ctx, state) when ctx.playback_state == :playing do
```
mickel8, line 88: maybe `{:frame_provider, message}` and then
`case message do`

or split this into multiple function clauses

lib/membrane_rtmp_plugin/rtmp/source.ex
```diff
@@ -58,10 +66,40 @@ defmodule Membrane.RTMP.Source do
 
   @impl true
   def handle_demand(pad, _size, _unit, _ctx, state) do
-    case Native.read_frame(state.native) do
-      {:ok, type, frame} ->
+    {actions, state} = maybe_provide_frame(pad, state)
```
mickel8, line 69: I would include `maybe_provide_frame` body here but up to you

lib/membrane_rtmp_plugin/rtmp/source.ex
```diff
@@ -71,6 +109,39 @@ defmodule Membrane.RTMP.Source do
...
+          :terminate ->
+            Native.stop(native)
+            :ok
+        end
+    end
+  end
+
+  @impl true
+  def handle_playing_to_prepared(_ctx, state) do
+    send(state.provider, :terminate)
```
mickel8, line 141: Here we should call `Native.stop`

lib/membrane_rtmp_plugin/rtmp/source.ex
```diff
@@ -58,10 +66,40 @@ defmodule Membrane.RTMP.Source do
...
-        {{:ok, buffer: {type, %Buffer{payload: payload}}, redemand: pad}, state}
+        buffer = %Buffer{pts: timestamp, metadata: %{timestamp: timestamp}, payload: payload}
+        state = update_in(state, [:buffers, type], &Qex.push(&1, buffer))
+
+        if Enum.count(get_in(state, [:buffers, type])) > state.max_buffer_size,
+          do: raise("Buffer limit exceeded on pad #{inspect(type)}. Check your demands")
+
+        {{:ok, []}, state}
+
+        {{:ok, redemand: other}, state}
```
mickel8, line 102: as we disscussed this doesn't work perfectly
also, the line above seem to be redundant

mickel8 (changes requested):

lib/membrane_rtmp_plugin/rtmp/source.ex
```diff
@@ -71,6 +109,39 @@ defmodule Membrane.RTMP.Source do
...
+            send(target, {:result, result})
+
+            if result == :end_of_stream, do: :ok, else: frame_provider(native, target)
+
+          :terminate ->
+            Native.stop(native)
+            :ok
+        end
+    end
+  end
```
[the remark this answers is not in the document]
daniel-jodlos: Second version doesn't have a priority for :terminate. If we request 1000 frames and then terminate, the element will still produce 1000 frames and then terminate

lib/membrane_rtmp_plugin/rtmp/source.ex
daniel-jodlos: Frame provider can be waiting for a frame right now. If we release the state, this might cause problems and is generally a race condition. I think it's better to hand the Native over to the frame provider. In fact, maybe we shouldn't even keep it in the state. Just initialize, create frame provider and forget

c_src/membrane_rtmp_plugin/rtmp.c
```diff
@@ -61,6 +61,11 @@ UNIFEX_TERM native_create(UnifexEnv *env, char *url, char *timeout) {
   return ret;
 }
 
+UNIFEX_TERM stop(UnifexEnv* env, State* s) {
+  unifex_release_state(env, s);
```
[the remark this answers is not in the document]
daniel-jodlos: I don't think so. I use it to ensure that the resources are terminated when the transition from playing to prepared occurs

mickel8 (approved):
Could we write somewhere how this demuxer works? I mean adjustment to slower demand

PR #8, opened 2021-12-15
Implement RTMPS client side (sink element).

marcinwasowicz:


daniel-jodlos (changes requested):

examples/stream.exs
```diff
@@ -0,0 +1,57 @@
+defmodule Stream do
```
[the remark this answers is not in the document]
marcinwasowicz: elixir examples/rtmp_sink.exs will indeed fail, but I launched it with mix run. Is it not recommended to use mix run?

mix.exs
```diff
@@ -42,11 +42,15 @@ defmodule Membrane.RTMP.Mixfile do
...
-      {:membrane_h264_ffmpeg_plugin, "~> 0.12"},
-      {:membrane_aac_plugin, "~> 0.9"},
-      {:ffmpex, "~> 0.7", only: :test}
+      {:membrane_file_plugin, "~> 0.6"},
+      {:membrane_aac_format, "~> 0.5", override: true},
+      {:membrane_element_fake, "~> 0.5"},
+      {:membrane_h264_ffmpeg_plugin, "~> 0.15"},
+      {:membrane_aac_plugin, "~> 0.11.0"},
+      {:ffmpex, "~> 0.7", only: :test},
+      {:membrane_realtimer_plugin, "~> 0.4.0"},
```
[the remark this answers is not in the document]
marcinwasowicz: I do not understand. Which particular plugin do you refer to?

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,142 @@
...
+    case Native.init_video_stream(state.native_state, caps.width, caps.height, caps.content.avcc) do
+      {:ok, ready, native_state} ->
+        Membrane.Logger.debug("Correctly initialized video stream.")
+        state = Map.put(state, :native_state, native_state)
+
+        if ready do
+          write_header(ctx, state)
+        else
+          {:ok, state}
+        end
```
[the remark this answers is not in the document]
marcinwasowicz: Actually it is more complex. Demanding frames and attempting to send them before writing a header will result in segmentation fault so I need to wait for both caps before demanding buffers from pads. This is what I use `ready` variable for. My suggestion is that I will remove writing header from elixir but I will still return `ready` variable from functions responsible for initializing audio and video streams.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,204 @@
...
+
+    if(state->video_stream_index == -1){
+        video_stream = avformat_new_stream(state->output_ctx, NULL);
+        if(!video_stream){
+            return init_video_stream_result_error(env, "Failed allocation video stream");
+        }
+        state->video_stream_index = video_stream->index;
+    }
+    else{
+        video_stream = state->output_ctx->streams[state->video_stream_index];
```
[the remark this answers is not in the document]
marcinwasowicz: Are we sure that receiving caps for the same stream several times is not expected to happen? If I raise here code will fail if caps for the same stream are sent more than once. I might also just return immediately if stream has been already initialized.

daniel-jodlos:
No, it isn’t there are two problems here:
- Examples aren’t a part of this mix project
- You add unnecessary dependencies to `mix.exs` which then make it into the released package

mix.exs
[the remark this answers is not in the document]
daniel-jodlos: The one that this comment refers to. The realtimer

examples/stream.exs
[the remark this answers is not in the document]
daniel-jodlos: No, it isn't. There are two problems here:
- This example is not a part of this mix project
- You are adding unnecessary stuff to mix.exs which will then make it into the released package

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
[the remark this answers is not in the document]
daniel-jodlos: I think I would just raise if any of the caps is `nil` and `handle_write/4` is called. What you are doing here is nice, but it uses some magic and the fact that you need these caps to be sent is not clear. Also, some elements, like H264 Parser or AAC Parser need to get a demand before they will generate any caps, because they rely on byte stream data. This Sink only works because Buffer between elements demands some data just to fill itself.

- Demand as always in `handle_prepared_to_playing`
- Raise if `handle_write` is called when any of the caps is missing. You can also check it in native code if you prefer, it doesn't matter to me
- If you do the above, you can just `if(ready) write_header` in native code wherever you calculate `ready` and remove `write_header` from the spec - it will actually be less complex, especially on Elixir side

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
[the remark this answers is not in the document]
daniel-jodlos: But can you even write the header twice? I doesn't seem very likely to me

examples/stream.exs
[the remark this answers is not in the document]
marcinwasowicz: I introduced required changes. Script can now be run directly with `elixir` command. I will put instruction hot to use it in README.

mix.exs
```diff
@@ -42,11 +42,15 @@ defmodule Membrane.RTMP.Mixfile do
       {:ex_doc, "~> 0.24", only: :dev, runtime: false},
       {:dialyxir, "~> 1.1", only: :dev, runtime: false},
       {:credo, "~> 1.5", runtime: false},
-      {:membrane_file_plugin, "~> 0.6", only: :test},
-      {:membrane_aac_format, "~> 0.3"},
-      {:membrane_h264_ffmpeg_plugin, "~> 0.12"},
-      {:membrane_aac_plugin, "~> 0.9"},
-      {:ffmpex, "~> 0.7", only: :test}
+      {:membrane_file_plugin, "~> 0.6"},
```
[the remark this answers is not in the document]
marcinwasowicz: I marked file plugin as test-only dependency.

mix.exs
```diff
@@ -42,11 +42,15 @@ defmodule Membrane.RTMP.Mixfile do
...
+      {:membrane_aac_format, "~> 0.5", override: true},
```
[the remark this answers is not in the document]
marcinwasowicz: Solved.

mix.exs
```diff
@@ -42,11 +42,15 @@ defmodule Membrane.RTMP.Mixfile do
...
+      {:membrane_element_fake, "~> 0.5"},
```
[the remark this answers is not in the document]
marcinwasowicz: Ok, I removed element fake dependency.

mix.exs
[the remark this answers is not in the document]
marcinwasowicz: Applied. Realtimer is now directly installed when running examples/stream.exs.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -0,0 +1,67 @@
...
+
+  alias Membrane.Testing
+  alias Membrane.Testing.{Pipeline}
+
+  @input_video_path "test/fixtures/bun33s_480x270.h264"
+  @input_audio_path "test/fixtures/bun33s.aac"
+
+  @flv_path "/tmp/rtmp_sink_test.flv"
+
+  test "Check if audio and video streams are correctly dumped to .flv file" do
```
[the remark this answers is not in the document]
marcinwasowicz: As required I changed it to test whether sink is correctly streaming to RTMP server launched via ffmpeg.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,142 @@
...
+    availability: :always,
+    caps: MP4.Payload,
+    mode: :pull,
+    demand_unit: :buffers
+
+  def_options rtmp_url: [
+                type: :string,
+                spec: String.t(),
+                description: "URL address of an RTMP/RTMPS stream destination (e.g. youtube)."
+              ]
```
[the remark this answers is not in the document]
marcinwasowicz: I understand that if `rtmp_url` parameter is a path to a file then this sink is effectively working as an FLV muxer, but I am rather against disabling this option.  Ability to dump muxed audio and video streams to a file that could be later investigated allowed us to detect and fix many basic errors without needing to launch additional RTMP server as debugger. It might be useful for developers that will work on this plugin in the future.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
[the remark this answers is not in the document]
marcinwasowicz: If I demand from both pads in `handle_prepared_to_playing` I may receive first video frame after receiving H264 caps, but before AAC caps. I spotted this case after getting segmentation fault from time to time at irregular rate because I was attempting to write frame prior to sending header. If I implement the steps above it will mean that I allow it to happen but return an informative error instead of segmentation fault. Do we accept this? Let me know if I understand you correctly.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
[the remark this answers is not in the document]
daniel-jodlos: This is not only a file problem. Someone might skip `rtmp://` part all together and who knows how FFmpeg will react. One thing I can tell you for sure, this will not be a pretty error and good luck finding this bug as someone who is just trying to stream. A lot of these guys don't have half the knowledge of the multimedia that you do and precisely 0 of that plugin and they will either give up or end up on our Discord with crazy errors. If someone knows that they can debug it like that, I think they can comment out a few lines of code.

I understand that the same exploit is technically possible with RTMP.Source, but the fact that I made a mistake there doesn't mean you need to duplicate it here.

If you really want to, you can leave a CODE comment (starting with `#`, not `@doc`) that they can comment out the check and save to an FLV file, stating it is not recommended, but I want it to be a conscious action and not the part of the public API.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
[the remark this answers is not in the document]
daniel-jodlos: No, we do not accept any failure that is caused by race condition in our code. Neither do we accept any magic. What you need to do here is buffer that problematic frame and send it ASAP. If you demand correctly, I think that this will only be one frame, if there is more, you can probably raise, but I'm not a 100% sure, think about it.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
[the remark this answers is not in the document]
marcinwasowicz: With new changes stream parameters are not overwritten by repeated caps (function returns immediately if stream is already initialized).

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,204 @@
+#include "rtmp_sink.h"
+#include <stdlib.h>
+
+const float H264_BPP = 0.116f;
+const AVRational MEMBRANE_TIME_BASE  = (AVRational) {1, 1000000000};
```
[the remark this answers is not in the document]
marcinwasowicz: H264_BPP was used to calculate bitrate of the stream upon video stream initialization. In fact this parameter is not required to write stream header and streaming to youtube/ffmpeg rtmp server was successful without its calculation. Therefore I removed this constant.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,204 @@
...
+    return close_connection_result_ok(env);
+}
+
+UNIFEX_TERM init_video_stream(UnifexEnv* env, State* state, int width, int height, UnifexPayload* avc_config){
+    AVStream* video_stream;
+
+    if(state->video_stream_index == -1){
+        video_stream = avformat_new_stream(state->output_ctx, NULL);
+        if(!video_stream){
+            return init_video_stream_result_error(env, "Failed allocation video stream");
```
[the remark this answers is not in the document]
marcinwasowicz: All malloc and stream initialization failures are now leading to `unifex_raise` call.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,204 @@
...
+
+    if (!(output_ctx->oformat->flags & AVFMT_NOFILE)) {
+        if(avio_open(&output_ctx->pb, rtmp_url, AVIO_FLAG_WRITE) < 0){
+            unifex_release_state(env, state);
+            return init_connection_result_error(env, "Failed to open provided URL for writing.");
+        }
+    }
+
+    state->output_ctx = output_ctx;
+    return init_connection_result_ok(env, state);
```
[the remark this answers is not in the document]
marcinwasowicz: If I release state before create_result_ok (which uses state) I will get segmentation fault. Perhaps you meant releasing state before returning error from NIF?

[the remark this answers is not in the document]
daniel-jodlos: I'm not sure if you will, but if you do, you can always do something like this:
```elixir
UNIFEX_TERM result = init_connection_resultt_ok(state);
unifex_release_state(env, state);
return result;
```
And I would strongly suggest using `.goto` I mentioned in another comment

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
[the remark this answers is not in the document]
daniel-jodlos: That's fine with me, as long as you remember to signal it to the end user somehow

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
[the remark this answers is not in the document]
marcinwasowicz: Now init_audio_stream and init_video_stream do not return error - they just return ok on success or raise. I suggest that we remove {:error :: label, reason :: string} from their spec.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,210 @@
...
+    if(is_key_frame){
+        packet->flags |= AV_PKT_FLAG_KEY;
+    }
+
+    packet->stream_index = state->video_stream_index;
+    packet->size = frame->size;
+    packet->data = (uint8_t*) av_malloc(frame->size);
+
+    if(!packet->data){
+        return unifex_raise(env, "Failed allocating video frame data");
```
daniel-jodlos, line 126: Memory leak!

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -0,0 +1,210 @@
...
+    packet->size = frame->size;
+
+    packet->data = (uint8_t*) av_malloc(frame->size);
+    if(!packet->data){
+        return unifex_raise(env, "Failed allocating audio frame data.");
+    }
+    memcpy(packet->data, frame->data, frame->size);
+    int64_t pts_scaled = av_rescale_q(pts, MEMBRANE_TIME_BASE, audio_stream_time_base);
+
+    packet->dts = pts_scaled;
```
daniel-jodlos, line 170: Leave a comment explaining why you set it

c_src/membrane_rtmp_plugin/sink/rtmp_sink.spec.exs
```diff
@@ -0,0 +1,13 @@
+module Membrane.RTMP.Sink.Native
+
+state_type "State"
+interface [NIF]
+
+spec create(rtmp_url :: string) :: {:ok :: label, state} | {:error :: label, reason :: string}
+spec close_stream(state) :: :ok
```
daniel-jodlos, line 7:
```suggestion
spec finalize_stream(state) :: (:ok :: label)
```

examples/stream.exs
```diff
@@ -0,0 +1,82 @@
+Mix.install([
+  {:membrane_realtimer_plugin, "~> 0.4.0"},
+  {:membrane_file_plugin, "~> 0.6"},
+  {:membrane_h264_ffmpeg_plugin, "~> 0.15.0"},
+  {:membrane_aac_plugin, "~> 0.11.0"},
+  {:membrane_mp4_plugin, github: "membraneframework/membrane_mp4_plugin"},
+  {:membrane_rtmp_plugin, path: __DIR__ |> Path.join("..") |> Path.expand()}
```
daniel-jodlos, line 7: - add core dependency
- use mp4 plugin v0.10.0
- use h264 plugin v0.16.1

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+    caps: MP4.Payload,
+    mode: :pull,
+    demand_unit: :buffers
+
+  def_options rtmp_url: [
+                type: :string,
+                spec: String.t(),
+                description:
+                  "Destination URL of the stream. It needs to start with rtmp:// or rtmps:// depending on the protocol variant.
+                This URL should be provided to you by your streaming service."
```
daniel-jodlos, line 32:
```suggestion
                This URL should be provided by your streaming service."
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+              ]
+
+  @impl true
+  def handle_init(options) do
+    case options.rtmp_url do
+      "rtmp://" <> _ -> {:ok, Map.from_struct(options)}
+      "rtmps://" <> _ -> {:ok, Map.from_struct(options)}
+      _url -> raise("Invalid destination URL provided")
```
daniel-jodlos, line 40: This needs some refactoring as we discussed

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+
+      {:error, reason} ->
+        raise("Transition to state playing failed with reason: #{reason}")
+    end
+  end
+
+  @impl true
+  def handle_playing_to_prepared(_ctx, state) do
+    Native.close_stream(state.native)
+    state = Map.drop(state, [:native, :ready, :buffered_frames])
```
daniel-jodlos, line 61: Assume that structure of state is contant

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+        _ctx,
+        state
+      ) do
+    case Native.init_video_stream(state.native, caps.width, caps.height, avc_config) do
+      {:ok, ready, native} ->
+        Membrane.Logger.debug("Correctly initialized video stream.")
+        state = Map.merge(state, %{native: native, ready: ready})
+        {:ok, state}
+
+      {:error, :caps_resend} ->
```
daniel-jodlos, line 79:
```suggestion
      {:error, :caps_resent} ->
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+        Membrane.Logger.debug("Re-received caps for video stream")
```
daniel-jodlos, line 80:
```suggestion
        Membrane.Logger.error("Input caps redefined on pad ... . RTMP Sink doesn't support changing parameters")
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+    case Native.write_audio_frame(state.native, buffer.payload, buffer_pts) do
+      {:ok, native} ->
+        state = Map.put(state, :native, native)
+
+        state =
+          Map.update(state, :current_timestamps, %{audio: buffer_pts}, fn curr_tmps ->
+            Map.put(curr_tmps, :audio, buffer_pts)
+          end)
+
+        {{:ok, demand: state.current_timestamps |> Enum.min_by(&elem(&1, 1)) |> elem(0)}, state}
```
daniel-jodlos, line 161: Extract demanding to other function

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,183 @@
...
+      {:error, reason} ->
+        raise("Writing audio frame failed with reason: #{reason}")
+    end
+  end
+
+  @impl true
+  def handle_end_of_stream(
+        pad,
+        _ctx,
+        %{current_timestamps: %{audio: _audio_dts, video: _video_dts} = curr_tmps} = state
```
daniel-jodlos, line 172: For detection of true end of stream, rely on context

mix.exs
```diff
@@ -43,10 +43,10 @@ defmodule Membrane.RTMP.Mixfile do
...
       {:credo, "~> 1.5", runtime: false},
       {:membrane_file_plugin, "~> 0.6", only: :test},
-      {:membrane_aac_format, "~> 0.3"},
-      {:membrane_h264_ffmpeg_plugin, "~> 0.12"},
-      {:membrane_aac_plugin, "~> 0.9"},
-      {:ffmpex, "~> 0.7", only: :test}
+      {:membrane_h264_ffmpeg_plugin, "~> 0.15.0"},
+      {:membrane_aac_plugin, "~> 0.11.0"},
+      {:ffmpex, "~> 0.7", only: :test},
+      {:membrane_mp4_plugin, github: "membraneframework/membrane_mp4_plugin"}
```
daniel-jodlos, line 49: Same as with the example

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -0,0 +1,110 @@
...
+
+  alias Membrane.Testing
+  alias Membrane.Testing.{Pipeline}
+
+  require Logger
+
+  @input_video_path "test/fixtures/bun33s_480x270.h264"
+  @input_audio_path "test/fixtures/bun33s.aac"
+
+  @rtmp_srv_url "rtmp://localhost:49500"
```
daniel-jodlos, line 13:
```suggestion
  @rtmp_server_url "rtmp://localhost:49500"
```

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -0,0 +1,110 @@
...
+    assert_pipeline_playback_changed(sink_pipeline_pid, :prepared, :playing)
+
+    assert_start_of_stream(sink_pipeline_pid, :rtmps_sink, :video, 5_000)
+    assert_start_of_stream(sink_pipeline_pid, :rtmps_sink, :audio, 5_000)
+
+    assert_end_of_stream(sink_pipeline_pid, :rtmps_sink, :video, 5_000)
+    assert_end_of_stream(sink_pipeline_pid, :rtmps_sink, :audio, 5_000)
+
+    Membrane.Testing.Pipeline.stop_and_terminate(sink_pipeline_pid, blocking?: true)
+    assert File.exists?(@flv_path)
```
daniel-jodlos, line 41: - Check if this file is not empty

daniel-jodlos (changes requested):

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,197 @@
...
+
+        {:ok, state}
+    end
+  end
+
+  @impl true
+  def handle_write(pad, buffer, _ctx, %{ready: false} = state) do
+    state = Map.put(state, :buffered_frame, {pad, buffer})
+    {:ok, state}
+  end
```
mat-hek, line 125: why is this needed?

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,197 @@
...
+      {:error, reason} ->
+        raise("Writing audio frame failed with reason: #{reason}")
+    end
+  end
+
+  @impl true
+  def handle_end_of_stream(
+        pad,
+        _ctx,
+        %{current_timestamps: %{audio: _audio_dts, video: _video_dts} = curr_tmps} = state
```
mat-hek, line 186: Don't use abbreviations when not necessary. Maybe moving that to the function body would be more readable

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,197 @@
...
+    case Native.write_audio_frame(state.native, buffer.payload, buffer_pts) do
+      {:ok, native} ->
+        state = Map.put(state, :native, native)
+
+        state =
+          Map.update!(state, :current_timestamps, fn curr_tmps ->
+            Map.put(curr_tmps, :audio, buffer_pts)
+          end)
+
+        {{:ok, demand: state.current_timestamps |> Enum.min_by(&elem(&1, 1)) |> elem(0)}, state}
```
mat-hek, line 175: Hard to grasp what's happening here. Generally, avoid using `elem/2`

README.md
```diff
@@ -90,7 +94,89 @@ To run this example you will need following extra dependency
...
+ffmpeg -listen 1 -f flv -i rtmp://localhost:1935 -c copy dest.flv
+```
+It will receive stream and once streaming is completed dump it to .flv file.
+Run it with:
+```elixir
+pipeline_options = %{
+  video_file_path: "bun33s_480x270.h264",
+  audio_file_path: "bun33s.aac",
+  rtmp_url: System.get_env("RTMP_URL")
+}
```
mat-hek, line 170: Options for an example pipeline are overkill. Especially when they're passed in different ways.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -0,0 +1,112 @@
+defmodule Membrane.RTMP.Sink.Test do
```
mat-hek, line 1: A test for file -> RTMP sink -> RTMP source -> file would be useful. I'd consider making it an example too.

mat-hek (changes requested):
looks great, some comments from me

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -0,0 +1,197 @@
...
+    case Native.init_video_stream(state.native, caps.width, caps.height, avc_config) do
+      {:ok, ready, native} ->
+        Membrane.Logger.debug("Correctly initialized video stream.")
+        state = Map.merge(state, %{native: native, ready: ready})
+        {:ok, state}
+
+      {:error, :caps_resent} ->
+        Membrane.Logger.error(
+          "Input caps redefined on pad :video. RTMP Sink does not support changing stream parameters"
+        )
```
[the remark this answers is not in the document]
marcinwasowicz: I was discussing it with @daniel-jodlos previously. RTMP Sink takes advantage of membrane H264 parser which is resending caps several times. It is a bug in parser implementation that is expected to be resolved and then we will be able to raise.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
marcinwasowicz: Together with @daniel-jodlos and @bblaszkow06 we agreed that we must demand from both audio and video pads in handle_prepared_to_playing to avoid "magic" in membrane. If I demand from both pads the following scenario can occur (and indeed occurs quite frequently): sink might receive first video frame before receiving audio caps. Unfortunately I cannot send any frame before I receive both caps. Bot caps are needed to send correct stream header. Therefore we agreed with @daniel-jodlos and @bblaszkow06 that any "falstart" frame mus be buffered and send ASAP once both caps are received and stream header is correctly sent.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
marcinwasowicz: This function clause will be entirely refactored (and this abbreviation will be removed) as requested by @daniel-jodlos.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
marcinwasowicz: As discussed today on meeting with @daniel-jodlos and @bblaszkow06 I will attempt to implement such an integration test. However in case of difficulties its implementation will be moved to a separate issue.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
mat-hek: So you can just wait until you get both caps and then make demands ;)

test/membrane_rtmp_plugin/rtmp_sink_test.exs
mat-hek: Cool, it can be a separate issue

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
marcinwasowicz: Before the first round of code review it was implemented this way to wait for both caps before any demands. @daniel-jodlos insisted that we demand from handle_prepared_to_playing. I will consult with him it once again.

mat-hek: cc @bblaszkow06 @daniel-jodlos

bblaszkow06: > So you can just wait until you get both caps and then make demands

The problem is that this works only thanks to input buffer making an initial demand - there's no guarantee given by the framework that caps will arrive without making a demand (even though that's what will happen) and I believe it's a kind of magic that we want to avoid in Membrane

mat-hek: The problem is that caps themselves are not being demanded, so from the sink perspective it's counter-intuitive that demand is required to receive them. I believe it should be the responsibility of each element to send caps and make a demand if it's needed to generate them. This is what we'd have to standardize, but it's more intuitive, easier to implement and potentially may speed up things, as caps could be generated before the sink requests any demands (even if no initial demand was generated)

README.md
```diff
@@ -4,9 +4,15 @@
...
 [![CircleCI](https://circleci.com/gh/membraneframework/membrane_rtmp_plugin.svg?style=svg)](https://circleci.com/gh/membraneframework/membrane_rtmp_plugin)
 
-This package provides RTMP server which listens to a connection from a client.
+This package provides RTMP server which listens to a connection from a client and element for streaming to an RTMP server.
+### Server
 After establishing connection it receives RTMP stream, demux it and outputs H264 video and AAC audio.
 At this moment only one client can connect to the server.
+### Streaming element
+After establishing connection with server it waits to receive video and audio streams. Once both streams are received they are streamed to the server.
+Currently the following codecs are supported:
```
daniel-jodlos, line 13:
```suggestion
Currently  only the following codecs are supported:
```

README.md
```diff
@@ -90,7 +97,23 @@ To run this example you will need following extra dependency
...
+```bash
+export RTMP_URL=rtmp://localhost:1935
+ffmpeg -listen 1 -f flv -i rtmp://localhost:1935 -c copy dest.flv
+```
+It will receive stream and once streaming is completed dump it to .flv file. If using the command above, please remember to run it **before** the streaming script.
+Running this example requires the following extra dependencies:
+```elixir
+{:membrane_realtimer_plugin, "~> 0.4.0"},
+{:membrane_hackney_plugin, "~> 0.6.0"},
+```
```
daniel-jodlos, line 116: Shouldn't be true. Mix.install should take care of all of that

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -12,6 +12,14 @@ defmodule Membrane.RTMP.Sink do
   alias Membrane.{AAC, MP4}
   require Membrane.Logger
 
+  @supported_protocols ["rtmp://", "rtmps://"]
+  @default_state_attributes %{
```
daniel-jodlos, line 16: How about simply
```suggestion
  @default_state %{
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -114,70 +128,63 @@ defmodule Membrane.RTMP.Sink do
...
+  def handle_write(pad, buffer, _ctx, state) do
+    state = write_frame(state, pad, buffer)
+    {{:ok, demand: get_demand(state)}, state}
+  end
 
-      {:error, reason} ->
-        raise("Writing video frame failed with reason: #{reason}")
+  @impl true
+  def handle_end_of_stream(pad, ctx, state) do
+    if ctx.pads |> Map.values() |> Enum.map(& &1.end_of_stream?) |> Enum.all?() do
```
daniel-jodlos, line 146: Does it work? I thought that context will be updated with eos for the current pad after this callback is called. Also, you overcomplicated a little bit
```suggestion
    if ctx.pads |> Map.delete(pad) |> Map.values() |> Enum.all?(& &1.end_of_stream?) do
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -114,70 +128,63 @@ defmodule Membrane.RTMP.Sink do
...
+      {:error, reason} ->
+        raise("Writing video frame failed with reason: #{reason}")
+    end
   end
 
-  @impl true
-  def handle_end_of_stream(_pad, _ctx, state) do
-    {:ok, Map.delete(state, :current_timestamps)}
+  defp get_demand(state) do
+    state.current_timestamps |> Enum.min_by(fn {_pad, timestamp} -> timestamp end) |> elem(0)
```
daniel-jodlos, line 188: Three things:
- avoid `elem/2` if possible
- timestamp can be `Ratio.t()`, and then it will be constantly demanding from one pad - you need to fix that. Also you can use `Bunch.value/1`, but I don't insist
- I'd rather you returned `{pad, 1}`. This will signal the intention a little more clearly

daniel-jodlos (changes requested):
Looks good, but I found some bugs in the code.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
marcinwasowicz: I make sure that timestamp of audio frame is integer at line 155 in sink.ex.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
marcinwasowicz: After running experiments with detailed logging it appears that int the case of handle_end_of_stream context is updated prior to callback execution. Nevertheless, I will remove unnecessary Enum.map as you suggested.

daniel-jodlos (approved):

mat-hek (approved):

PR #12, opened 2022-05-16
Update dependencies so that to support membrane core 0.10.0

varsill:
Update dependencies so that to support membrane core 0.10.0


mat-hek (changes requested):
please update unifex as well

PR #18, opened 2022-05-30
Update repo to current plugin template

bblaszkow06:

PR #19, opened 2022-06-03
Fix sink test reliability - an option to retry connection

bblaszkow06:
Should be merged after #18

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
@@ -15,7 +15,9 @@ defmodule Membrane.RTMP.Sink do
   alias Membrane.{AAC, MP4}
 
   @supported_protocols ["rtmp://", "rtmps://"]
+  @connection_attempt_interval 500
   @default_state %{
+    attempts: 0,
     native: nil,
     buffered_frame: nil,
     ready: false,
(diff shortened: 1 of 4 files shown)


lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -40,34 +42,40 @@ defmodule Membrane.RTMP.Sink do
                 description:
                   "Destination URL of the stream. It needs to start with rtmp:// or rtmps:// depending on the protocol variant.
                 This URL should be provided by your streaming service."
+              ],
+              max_attempts: [
+                type: :integer,
+                spec: pos_integer(),
+                default: 1,
```
daniel-jodlos, line 49: I'm wondering if it shouldn't be set to a higher value 🤔

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -40,34 +42,40 @@ defmodule Membrane.RTMP.Sink do
...
-      raise("Invalid destination URL provided")
+    unless String.starts_with?(options.rtmp_url, @supported_protocols) do
+      raise ArgumentError, "Invalid destination URL provided"
+    end
+
+    unless is_integer(options.max_attempts) and options.max_attempts >= 1 do
+      raise ArgumentError, "Invalid max_attempts option value: #{options.max_attempts}"
     end
+
+    {:ok, Map.from_struct(options) |> Map.merge(@default_state)}
```
daniel-jodlos, line 66:
```suggestion
    {:ok, options |> Map.from_struct() |> Map.merge(@default_state)}
```

daniel-jodlos (approved):

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
bblaszkow06: I wanted to keep the default behaviour. Besides, it won't always be needed and I think making the retry mechanism should stay optional.

PR #20, opened 2022-06-03
Prevent Source from blocking a port forever

bblaszkow06:
Should be merged after #18

Fixes a serious problem where `RTMP.Source` would listen forever for a connection, essentially disabling the port it is using until the VM is restarted

### DIFF
--- a/c_src/membrane_rtmp_plugin/source/rtmp_source.c
+++ b/c_src/membrane_rtmp_plugin/source/rtmp_source.c
@@ -3,7 +3,13 @@
 
 void handle_init_state(State *);
 
-UNIFEX_TERM native_create(UnifexEnv *env, char *url, char *timeout) {
+static int interrupt_cb(void *ctx) {
+  bool is_terminating = *(bool *)ctx;
+  // interrupt if the flag is set
+  return is_terminating;
+}
+
+UNIFEX_TERM create(UnifexEnv *env) {
   State *s = unifex_alloc_state(env);
   handle_init_state(s);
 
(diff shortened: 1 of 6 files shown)


lib/membrane_rtmp_plugin/rtmp/source/native.ex
```diff
@@ -4,27 +4,33 @@ defmodule Membrane.RTMP.Source.Native do
...
+  end
 
-  defp get_timeout(time),
-    do:
-      Time.as_seconds(time)
-      |> Ratio.trunc()
-      |> inspect()
-      |> then(&{:ok, &1})
+  defp get_int_timeout(time) do
+    Time.as_seconds(time) |> Ratio.trunc()
```
daniel-jodlos, line 34: Looks a bit weird
```suggestion
    time |> Time.as_seconds() |> Ratio.trunc()
```

c_src/membrane_rtmp_plugin/source/rtmp_source.c
```diff
@@ -3,7 +3,13 @@
 
 void handle_init_state(State *);
 
-UNIFEX_TERM native_create(UnifexEnv *env, char *url, char *timeout) {
+static int interrupt_cb(void *ctx) {
```
daniel-jodlos, line 6: Full name perhaps?
```suggestion
static int interrupt_callback(void *ctx) {
```

daniel-jodlos (approved):

c_src/membrane_rtmp_plugin/source/rtmp_source.c
Qizot: I believe that is the ffmpeg's convention. Not sure if we should follow it or not though.

bblaszkow06: Let's call a spade a spade - it's not a result of following the convention, it's just a copy-paste from ffmpeg's example :upside_down_face: And we definitely should follow more Elixir-like naming convention :wink:

PR #21, opened 2022-06-07
Prevent connection awaiting from blocking Source's process

bblaszkow06:
This should allow to terminate the pipeline gracefully when connection
listening has to be interrupted


Qizot:
The code looks fine but there is probably one use case missing. 

Currently, we are handling a case where the parent process exits then we shutdown the native process. There is a case where the native rtmp listens for a connection and the socket gets closed (before the media starts to flow), this results in a raise in the `await_connection` function (which is controlled by us), is the intended behavior or should we just stop the element as with do with the former case?

Qizot (approved):

PR #25, opened 2022-09-06
Created tcp_server 

shuntrho:


lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -0,0 +1,35 @@
...
+  """
+  use Task
+
+  @enforce_keys [:port, :tcp_options, :serve_fn]
+
+  defstruct @enforce_keys
+
+  @spec start_link(%__MODULE__{}) :: {:ok, pid}
+  def start_link(options) do
+    Task.start_link(__MODULE__, :run, [Enum.into(options, %{})])
```
Qizot, line 13:
```suggestion
    Task.start_link(__MODULE__, :run, [Map.from_struct(options)])
```

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -0,0 +1,35 @@
...
+  @moduledoc """
+  A simple tcp server, which executes given function for each new connection.
+  """
+  use Task
+
+  @enforce_keys [:port, :tcp_options, :serve_fn]
+
+  defstruct @enforce_keys
+
+  @spec start_link(%__MODULE__{}) :: {:ok, pid}
```
Qizot, line 11:
```suggestion
  @type t  :: %__MODULE__{
    port: :inet.port_number(),
    options: [:inet.inet_backend() | :gen_tcp.listen_option()],
    socker_handler:  (:gen_tcp.socket() -> {:ok, pid} | {:error, reason :: any()})
  }
  
  @spec start_link(t()) :: {:ok, pid}
```

In general, try to put as much specs as you can.

Also the module's doc could be more explanatory, for instance what the `serve_fn` is supposed to be used for (also the name could be changed to e.g. `socker_handler` or something).

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -0,0 +1,35 @@
...
+  @spec run(map()) :: nil
+  def run(options) do
+    {:ok, socket} = :gen_tcp.listen(options.port, options.tcp_options)
+
+    accept_loop(socket, options.serve_fn)
+  end
+
+  defp accept_loop(socket, serve_fn) do
+    {:ok, client} = :gen_tcp.accept(socket)
+    {:ok, pid} = Task.start(fn -> serve(client, serve_fn) end)
```
Qizot, line 25: I'm not sure if that logic is correct. The `serve` will spawn a process and try to give it a tcp control. It may happen that the `accept_loop` will do it later than the `serve` and you will have a race condition (serve can't give control unless itself being in control). You could spawn the serve task, give it control and then send it a message while `serve` will be waiting for the message and will continue with its logic.

test/membrane_rtmp_plugin/tcp_server_test.exs
```diff
@@ -0,0 +1,52 @@
+defmodule Membrane.RTMP.Source.TcpServer.Test do
```
Qizot, line 1:
```suggestion
defmodule Membrane.RTMP.Source.TcpServerTest do
```

test/membrane_rtmp_plugin/tcp_server_test.exs
```diff
@@ -0,0 +1,52 @@
...
+        :binary,
+        packet: :raw,
+        active: false,
+        reuseaddr: true,
+        ip: @local_ip
+      ],
+      serve_fn: fn socket ->
+        {:ok, receive_task} =
+          Task.start(fn ->
+            Process.whereis(__MODULE__) |> Process.link()
```
Qizot, line 23: Why not just save the pid as `caller = self()` before `server_options` instead of doing this whole registering? Also, why do you need to link the process?

Qizot (changes requested):

daniel-jodlos (approved):
I don't think this needs a Code Owner review, so you're getting a "wildcard approve" here

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
shuntrho: Yeah, you are right - race condition takes place here.

test/membrane_rtmp_plugin/tcp_server_test.exs
shuntrho: I registered the process probably for some debugging. I link the processes in order for the test to crash if the receiving process crashes.

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
shuntrho: I renamed the `serve_fn` and added example.

Qizot (approved):
Just update the moduledoc and otherwise LGTM

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -0,0 +1,88 @@
+defmodule Membrane.RTMP.Source.TcpServer do
+  @moduledoc """
+  A simple tcp server, which executes given function for each new incoming connection.
+
+  ## Example
+  ```
+  alias Membrane.RTMP.Source.TcpServer
```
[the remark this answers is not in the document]
shuntrho: done

PR #26, opened 2022-09-08
RTMP source pure elixir implementation

shuntrho:


lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -1,9 +1,10 @@
 defmodule Membrane.RTMP.SourceBin do
   @moduledoc """
-  Bin responsible for spawning new RTMP server.
+  Bin responsible for demuxing and parsing a video stream.
```
Qizot, line 3: I would add `parsing an RTMP stream`. Also it doesn't have to be video only, it can be both audio video or audio alone.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -1,9 +1,10 @@
...
 
-  It will receive RTMP stream from the client, parse it and demux it, outputting single audio and video which are ready for further processing with Membrane Elements.
-  At this moment only AAC and H264 codecs are support
+  The bin requires the RTMP client to be already connected.
+  Outputs single audio and video which are ready for further processing with Membrane Elements.
```
Qizot, line 6: The question probably could be if we intend to to support streams that can contains just video or just audio. OBS may always send the two but it won't be always the case anyways.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -21,34 +22,18 @@ defmodule Membrane.RTMP.SourceBin do
...
 
-  def_options port: [
-                spec: 1..65_535,
-                description: "Port on which the server will listen"
-              ],
-              local_ip: [
-                spec: binary(),
-                default: "127.0.0.1",
+  def_options socket: [
+                spec: port(),
```
Qizot, line 26: This probably should have the tcp socket spec. With socket backend for TCP sockets the implementation is no longer using ports and here you will receive some opaque tuple instead.

lib/membrane_rtmp_plugin/rtmp/source/interceptor.ex
```diff
@@ -0,0 +1,241 @@
+defmodule Membrane.RTMP.Interceptor do
+  @moduledoc """
+  #{inspect(__MODULE__)} is responsible for parsing
```
Qizot, line 3: I think that the meaning of the module has changed a little bit. It is always parsing messages since it returns the media packets as well. Its initial purpose was to stop intercepting once a certain message has been reached.

Also if you have some better name for it be free to suggest it 😉

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,264 @@
...
+  end
+
+  defp request_packet(pid) do
+    send(pid, :need_more_data)
+  end
+
+  defp get_media_buffers(header, data, state) do
+    payload =
+      case state.header_sent? do
+        false ->
```
Qizot, line 186: `if` instead of `case`

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,264 @@
...
+          get_flv_header()
+
+        true ->
+          <<>>
+      end <> get_flv_body(header, data)
+
+    buffers = [%Buffer{payload: payload} | state[:buffers]]
```
Qizot, line 193:
```suggestion
    buffers = [%Buffer{payload: payload} | state.buffers]
```

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,264 @@
...
+    {buffers, %{state | header_sent?: true}}
+  end
+
+  defp get_flv_header() do
+    <<"FLV", 0x01::8, 0::5, 1::1, 0::1, 1::1, 9::32>>
+  end
+
+  defp get_flv_body(
+         %Membrane.RTMP.Header{
+           #  chunk_stream_id: stream_id,
```
Qizot, line 203: delete if not needed

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,264 @@
...
+
+    payload = Message.chunk_payload(body, chunk_stream_id, chunk_size)
+
+    :gen_tcp.send(socket, [header | payload])
+  end
+
+  # The RTMP connection is based on TCP therefore we are operating on a continuous stream of bytes.
+  # In such case packets received on TCP sockets may contain a partial RTMP packet or several full packets.
+  #
+  # `Interceptor` is already able to request more data if packet is incomplete but it is not aware
```
Qizot, line 235: I'm wondering if it wouldn't be better if we operated on `recv` instead of the active pooling of messages. The we wouldn't have to care about those `need_more_data` and we would request as many bytes as we need.

But that shouldn't be done in that PR, rather a consideration for further improvement in the future.

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,84 @@
+defmodule Membrane.RTMP.SourceBin.Test do
+  use ExUnit.Case
```
Qizot, line 2: New line between use and import

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,84 @@
+defmodule Membrane.RTMP.SourceBin.Test do
```
Qizot, line 1: Also the test somehow should be marked as an integration test since we are using the external ffmpeg commands/

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,84 @@
...
+        Testing.Pipeline.start_link(options)
+      end
+    }
+
+    {:ok, _tcp_server} = TcpServer.start_link(options)
+    ffmpeg_task = Task.async(&start_ffmpeg/0)
+
+    pipeline = await_pipeline_started()
+
+    assert_pipeline_playback_changed(pipeline, :prepared, :playing)
```
Qizot, line 42: Would be great if you could know the duration of `testsrc.fflv` and make sure that you receive all membrane buffers representing audio and video (assert on them in a loop, make sure the timestamps are ok etc.).

test/support/rtmp_source_test_pipeline.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Membrane.RTMP.Source.Test.Pipeline do
```
Qizot, line 1:
```suggestion
defmodule Membrane.RTMP.Source.TestPipeline do
```

test/support/rtmp_source_test_pipeline.ex
```diff
@@ -0,0 +1,49 @@
...
+  @impl true
+  def handle_notification({:rtmp_source_initialized, source}, :src, _ctx, state) do
+    send(self(), {:rtmp_source_initialized, source})
+
+    {:ok, state}
+  end
+
+  @impl true
+  def handle_other({:rtmp_source_initialized, source}, _ctx, state) do
+    case :gen_tcp.controlling_process(state.socket, source) do
```
Qizot, line 39: I think we need a helper function inside the `SourceBin` that will hide the the raw `:gen_tcp.controlling_process`, it should just take the source payload and look something like 
```elixir
Membrane.RTMP.SourceBin.pass_control(state.socket, source)
```
Would be probably more friendly for a user than having to do that erlang style.

Qizot: Or have some other better mechanism, not sure how it should look though.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
Qizot: Also we are missing the documentation of what will happen with a socket and that a user will receive a notification that he needs to pass the control, information such as the socket should not be active and that the source will be the first one to request the packet.

Qizot (changes requested):
And one BIG thing that we are missing is the ability to authorize the stream based on the stream_key.

We need some behaviour for a module that will be passed to the SourceBin that will have callbacks such as reacting on the `Publish` and potentially allow the stream (authorization) and that will react on `SetDataFrame` as we may potentially want to ignore invalid stream parameters such as to big bitrate or something like that.

lib/membrane_rtmp_plugin/rtmp/source/amf_encoder.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  defp encode_key_value_pair({key, value}) do
+    [<<byte_size(key)::16>>, key, do_encode_object(value)]
+  end
+
+  @doc """
+  Helper function for debugging binaries by changing them to hex format
+  """
+  @spec to_hex(binary()) :: String.t()
+  def to_hex(message) do
```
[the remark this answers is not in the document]
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/interceptor.ex
```diff
@@ -0,0 +1,241 @@
...
+
+    {tx_ids, %{interceptor | current_tx_id: tx_id + n}}
+  end
+
+  @spec handle_packet(packet_t(), t()) ::
+          {Handshake.Step.t() | :need_more_data | :handshake_done | binary(), t()}
+          | {Header.t(), Message.t(), t()}
+  def handle_packet(packet, state)
+
+  # def handle_packet(packet, %{state_machine: :connected, buffer: <<>>} = state) do
```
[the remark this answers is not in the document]
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,264 @@
...
+
+  defp do_handle_client_message(%Messages.Publish{publish_type: _publish_type}, _header, _state) do
+    # NOTE: For now we end parsing messages once we receive publish message.
+    # At some point we may want to verify the stream parameters or simply extract
+    # them for further processing.
+    {:halt, {:error, :invalid_publish_type}}
+  end
+
+  defp do_handle_client_message(%Messages.SetDataFrame{}, _header, state) do
+    # raise "Received @setDataFrame RTMP message when it should not be possible"
```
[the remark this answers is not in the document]
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
shuntrho: done

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
shuntrho: done

test/support/rtmp_source_test_pipeline.ex
shuntrho: done

lib/membrane_rtmp_plugin/rtmp/source/interceptor.ex
shuntrho: I think `MessageParser` would be good.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
shuntrho: We want to stream just audio or video, and that should be possible now.

test/support/rtmp_source_test_pipeline.ex
shuntrho: We will stick with the basic wrapper for `:gen_tcp:controlling_process`.

lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -1,148 +1,155 @@
...
-      send(state.provider, :get_frame)
-      {{:ok, buffer: {type, buffer}}, state}
-    else
-      # if there is no demand for element of this type so we wait until it appears
-      # effectively, it results in source adapting to the slower of the two outputs
-      {:ok, %{state | stale_frame: {type, buffer}}}
-    end
+  def handle_playing_to_prepared(_ctx, state) do
+    send(state.receiver_pid, :terminate)
+    Process.unlink(state.receiver_pid)
```
Qizot, line 106: You have to check that out but from what I remember, if a process exits with a reason normal the links won't have any effect so you don't need to unlink then (unless you expect the receiver to live for a while_.

lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -1,148 +1,155 @@
...
-    Membrane.Logger.debug("Received end of stream")
-    {{:ok, end_of_stream: :audio, end_of_stream: :video}, state}
+  def handle_other(:start_receiving, _ctx, state) do
+    case :gen_tcp.controlling_process(state.socket, state.receiver_pid) do
+      :ok ->
+        :ok = :inet.setopts(state.socket, active: :once)
+        {:ok, %{state | socket_ready?: true}}
+
+      {:error, :not_owner} ->
+        Process.send_after(self(), :start_receiving, 200)
```
Qizot, line 118: Is it possible that we will never become the controlling process therefore we will be stuck in this function? We should probably give up trying after the first retry.

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Membrane.RTMP.Source.TcpServer do
   @moduledoc """
-  A simple tcp server, which executes given function for each new incoming connection.
+  A simple tcp server, which handles each new incoming connection.
 
   The `socket_handler` function passed inside the options should take the socket returned by `:gen_tcp.accept/1` and return `{:ok, pid}`, where the `pid` describes a process, which will be interacting with the socket. `Membrane.RTMP.Source.TcpServer` will grant that process control over the socket via `:gen_tcp.controlling_process/2`.
```
Qizot, line 5: Again, the line is too long.

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,119 @@
...
+
+    assert_buffers(%{
+      pipeline: pipeline,
+      sink: :video_sink,
+      stream_length: 3000,
+      buffers_expected: div(3000, 42)
```
Qizot, line 50: Can you make that `42` and `23` below module attributes as something like this `@video_frame_duration_ms` and `audio_frame_duration_ms`?

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,119 @@
...
+    })
+
+    assert_buffers(%{
+      pipeline: pipeline,
+      sink: :audio_sink,
+      stream_length: 3000,
+      buffers_expected: div(3000, 23)
+    })
+
+    assert_end_of_stream(pipeline, :audio_sink, :input, 11_000)
```
Qizot, line 60: why `11_000` specifically?

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -0,0 +1,119 @@
...
+      assert state.buffers >= state.buffers_expected
+    end
+  end
+
+  defp assert_buffers(state) do
+    stream_length = Membrane.Time.milliseconds(state.stream_length)
+
+    Map.merge(%{state | stream_length: stream_length}, %{last_dts: -1, buffers: 0})
+    |> assert_buffers()
+  end
```
Qizot, line 118:
```suggestion
    state
    |> Map.merge(%{stream_length: stream_length, last_dts: -1, buffers: 0})
    |> assert_buffers()
  end
```

test/support/rtmp_source_test_pipeline.ex
```diff
@@ -0,0 +1,56 @@
...
+  alias Membrane.RTMP.SourceBin
+  alias Membrane.Testing
+
+  @impl true
+  def handle_init(socket: socket, test_process: test_process) do
+    spec = %Membrane.ParentSpec{
+      children: [
+        src: %SourceBin{
+          socket: socket,
+          validator: Membrane.RTMP.Source.TestValidator
```
Qizot, line 16: I think that we should probably use `Verifier` instead of `Validator`, validation is usually about the data format which in our case is alright, we need to verify if we can let the stream to continue with given set of parameters.

Qizot: Also, can you create a test of what happens if the `Verifier` denies a particular stream? Also we can probably send a notification to the pipeline that a stream has been denied or something.

Qizot (changes requested):

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -1,9 +1,15 @@
 defmodule Membrane.RTMP.SourceBin do
   @moduledoc """
-  Bin responsible for spawning new RTMP server.
+  Bin responsible for demuxing and parsing an RTMP stream.
```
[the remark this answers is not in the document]
shuntrho: Technically, in the pipeline we first have demuxer, and then audio and video parsers.

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
shuntrho: The previous length of the stream was 10s i guess.

lib/membrane_rtmp_plugin/rtmp/source/source.ex
shuntrho: Ok, with retrying after 200ms a single retry wasn't enough to always succeed - I decided for 3 retries.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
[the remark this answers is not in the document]
Qizot: It's rtmp parsing, demuxing, video and audio parsing. So technically you are first parsing RTMP and then demuxing RTMP.

test/support/test_verifier.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Membrane.RTMP.Source.TestVerifier do
+  @moduledoc false
+  use Membrane.RTMP.MessageValidator
+
+  alias Membrane.RTMP.Messages
+
+  @stream_key "ala2137"
```
Qizot, line 7: 😉

lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -1,148 +1,161 @@
...
-
-                Duration given must be a multiply of one second or atom `:infinity`.
+                A Module implementing `Membrane.RTMP.MessageValidator` behaviour, used for validating the stream.
                 """
               ]
 
+  @typedoc """
+  Notification sent when the RTMP Source element is initialized and it should be granted control over the socket using `:gen_tcp.controlling_process/2`.
+  """
+  @type rtmp_source_initialized_t() :: {:rtmp_source_initialized, :gen_tcp.socket(), pid()}
```
Qizot, line 35: I think that we are missing other notifications here. I would also mention those types in the `moduledoc` as this is the first thing that people will read and they may skip some types that are not linked anywhere.

Qizot: Also I'm not sure if we need the `rtmp_` prefixes in notifications. Any other bin/pipeline will know that source of the notification.

Another thing, how about this particular notification being named `socket_control_needed`?

Qizot (changes requested):

Qizot (approved):
2 comments left, besides that LGTM, good job!

lib/membrane_rtmp_plugin/rtmp/source/tcp_server.ex
```diff
@@ -55,6 +55,7 @@ defmodule Membrane.RTMP.Source.TcpServer do
     receive do
       :control_granted ->
         :ok = :gen_tcp.controlling_process(socket, pid)
+        IO.inspect("granted control")
```
Qizot, line 58: Cool 😉

README.md
```diff
@@ -22,37 +14,70 @@ The package can be installed by adding `membrane_rtmp_plugin` to your list of de
...
+- AAC for audio
+
 ### Prerequisites
 In order to successfully build and install the plugin, you need to have **ffmpeg == 4.4** installed on your system
 
 ## Usage
-### `Membrane.RTMP.Source` - RTMP Server
+In order to run the examples make sure, you have installed the following plugins in your dependencies:
+
+```elixir
```
Qizot, line 42: Why did we remove the `Mix.install` form the example?

Qizot (commented):
There is also `received.flv` file committed to the repository.

lib/membrane_rtmp_plugin/rtmp/source/messages/command/connect.ex
```diff
@@ -0,0 +1,66 @@
+defmodule Membrane.RTMP.Messages.Connect do
+  @moduledoc false
+
+  @behaviour Membrane.RTMP.Message
+
+  alias Membrane.RTMP.AMFEncoder
+
+  @enforce_keys ~w(app type supports_go_away flash_version swf_url tc_url)a
+  defstruct [tx_id: 0] ++ @enforce_keys
```
daniel-jodlos, line 9: There were some comments about this already, please make sure that correct style applies everywhere

daniel-jodlos (changes requested):

examples/source.exs
```diff
@@ -3,61 +3,116 @@
...
 
-monitor_ref = Process.monitor(pid)
+    Membrane.RTMP.Source.TcpServer.start_link(server_options)
 
-# Wait for the pipeline to terminate
-receive do
-  {:DOWN, ^monitor_ref, :process, _pid, _reason} ->
-    :ok
+    receive do
+      :pipeline_terminated ->
```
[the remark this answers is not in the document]
shuntrho: I am not directly starting the pipeline here, so I do not have access to its pid.

lib/membrane_rtmp_plugin/rtmp/source/amf_parser.ex
```diff
@@ -0,0 +1,69 @@
...
+  defp parse_value(<<0x00, number::float-size(64), rest::binary>>) do
+    {number, rest}
+  end
+
+  # parsing a boolean
+  defp parse_value(<<0x01, boolean::8, rest::binary>>) do
+    case boolean do
+      0 -> {true, rest}
+      1 -> {false, rest}
+    end
```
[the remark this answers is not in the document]
shuntrho: nice

lib/membrane_rtmp_plugin/rtmp/source/header.ex
```diff
@@ -0,0 +1,163 @@
...
+  * `0b01` - current header derives the `stream_id` from the previous header
+  * `0b10` - same as above plus derives `type_id` and `body_size`
+  * `0b11` - all values are derived from the previous header with the same `chunk_stream_id`
+  """
+  @spec deserialize(binary(), t() | nil) :: {t(), rest :: binary()} | :need_more_data | :error
+  def deserialize(binary, previous_headers \\ nil)
+
+  # only the deserialization of the 0b00 type can have `nil` previous header
+  def deserialize(
+        <<@header_type_0::bitstring, chunk_stream_id::6, timestamp::24, body_size::24, type_id::8,
```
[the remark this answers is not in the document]
shuntrho: I prefer to add a little bit more code, but make it super explicit as to what header we are parsing in the clause.

lib/membrane_rtmp_plugin/rtmp/source/header.ex
```diff
@@ -0,0 +1,163 @@
...
+  """
+  @type t :: %__MODULE__{
+          chunk_stream_id: integer(),
+          timestamp: integer(),
+          body_size: integer(),
+          type_id: integer(),
+          stream_id: integer()
+        }
+
+  defmacro type(:set_chunk_size), do: 0x01
```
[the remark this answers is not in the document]
shuntrho: It is in the Message module.

examples/source.exs
```diff
@@ -3,61 +3,116 @@
...
+
+    {:ok, state}
   end
 
   # The rest of the module is used for self-termination of the pipeline after processing finishes
   @impl true
   def handle_element_end_of_stream({:sink, _pad}, _ctx, state) do
     Membrane.Pipeline.terminate(self())
+    send(state.parent, :pipeline_terminated)
     {{:ok, playback: :stopped}, state}
```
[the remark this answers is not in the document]
daniel-jodlos: 1. Parent should monitor the Pipeline, every other script does it this way
2. terminate actually transitions the state to `:stopped` for you, so there is no need to do this explicitly here

examples/source.exs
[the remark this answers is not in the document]
daniel-jodlos: Ok, fair enough. I think I missed that.

lib/membrane_rtmp_plugin/rtmp/source/message.ex
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def chunk_payload(payload, chunk_stream_id, chunk_size),
+    do: do_chunk_payload(payload, chunk_stream_id, chunk_size, [])
+
+  defp do_chunk_payload(payload, chunk_stream_id, chunk_size, acc)
+       when byte_size(payload) > chunk_size do
+    <<chunk::binary-size(chunk_size), rest::binary>> = payload
+
+    acc = [<<0b11::2, chunk_stream_id::6>>, chunk | acc]
```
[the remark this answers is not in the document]
Qizot: Nope, it is on purpose. The first part that you see in accumulator `<<0b11::2, chunk_stream_id::6>>` is a separator between the `chunk` and the `rest` payloads. So in the end it should appear after the chunk.

lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex
```diff
@@ -0,0 +1,231 @@
...
+  @enforce_keys [:state_machine, :buffer, :chunk_size, :handshake]
+  defstruct @enforce_keys ++ [previous_headers: %{}, current_tx_id: 1]
+
+  @type state_machine_t ::
+          :handshake | :connecting | :connected
+
+  @type packet_t :: binary()
+
+  @type t :: %__MODULE__{
+          state_machine: state_machine_t(),
```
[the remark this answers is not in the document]
Qizot: How do you define if it is a state machine or not? It holds atoms that are basically a continuation of the previous ones (just like in state machines).

If we have to change the name then I would not go for `state` as it mostly refers to a bunch of data which the `state_machine` is already a part of.

lib/membrane_rtmp_plugin/rtmp/source/amf_parser.ex
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  # parsing a string
+  defp parse_value(<<0x02, size::16, rest::binary>>) do
+    <<string::binary-size(size), rest::binary>> = rest
```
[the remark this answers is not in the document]
shuntrho: This situtation is unlikely if not impossible, as we are assuring that all message data is present before we parse it.

lib/membrane_rtmp_plugin/rtmp/source/amf_parser.ex
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  defp parse_object_pairs(
+         <<key_size::16, key::binary-size(key_size), rest::binary>>,
+         acc
+       ) do
+    {value, rest} = parse_value(rest)
+
+    parse_object_pairs(rest, [{key, value} | acc])
+  end
```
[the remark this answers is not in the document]
shuntrho: The objects will be complete, because we are only parsing a message when we have all the data.

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -21,34 +34,27 @@ defmodule Membrane.RTMP.SourceBin do
...
-  def handle_init(%__MODULE__{} = options) do
-    url = "rtmp://#{options.local_ip}:#{options.port}"
-    source = %RTMP.Source{url: url, timeout: options.timeout}
-
+  def handle_init(%__MODULE__{} = opts) do
     spec = %ParentSpec{
       children: %{
-        src: source,
+        src: %RTMP.Source{socket: opts.socket, validator: opts.validator},
+        demuxer: Membrane.FLV.Demuxer,
```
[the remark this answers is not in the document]
shuntrho: So far it's looking good.

lib/membrane_rtmp_plugin/rtmp/source/message.ex
```diff
@@ -0,0 +1,104 @@
+defmodule Membrane.RTMP.Message do
+  @moduledoc false
+
+  require Membrane.RTMP.Header
+
+  alias Membrane.RTMP.{Header, Messages}
+
+  @type message_data_t :: map() | number() | String.t() | :null
+
+  @type t :: struct()
```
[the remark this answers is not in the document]
shuntrho: Well, not really. Unless we want to specify all possible message types here.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,267 @@
...
+    payload =
+      get_flv_body(rtmp_header, data)
+      |> (&if(state.header_sent?, do: &1, else: get_flv_header() <> &1)).()
+
+    buffers = [%Buffer{payload: payload} | state.buffers]
+    %{state | header_sent?: true, buffers: buffers}
+  end
+
+  defp get_flv_header() do
+    <<"FLV", 0x01::8, 0::5, 1::1, 0::1, 1::1, 9::32>>
```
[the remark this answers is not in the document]
shuntrho: The thing is, that FLV sent inside the RTMP packets is partial, so we add the header at the beginning, so the FLV parser works correctly.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,267 @@
...
+  end
+
+  defp get_flv_body(
```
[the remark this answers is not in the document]
shuntrho: According to the documentation https://rtmp.veriskope.com/pdf/video_file_format_spec_v10.pdf everything after the header is called FLV file body.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,267 @@
...
+  end
+
+  defp do_handle_client_message(%Handshake.Step{type: :s0_s1_s2} = step, _header, state) do
+    :gen_tcp.send(state.socket, Handshake.Step.serialize(step))
+
+    connection_epoch = Handshake.Step.epoch(step)
+    {:cont, %{state | epoch: connection_epoch}}
+  end
+
+  defp do_handle_client_message(%Messages.SetChunkSize{chunk_size: _chunk_size}, _header, state) do
```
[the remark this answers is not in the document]
shuntrho: Yeah, it was a bug actually, thanks!

lib/membrane_rtmp_plugin/rtmp/source/bin.ex
```diff
@@ -60,11 +66,49 @@ defmodule Membrane.RTMP.SourceBin do
...
+        #
+        link(:demuxer)
+        |> via_out(Pad.ref(:audio, 0))
+        |> to(:audio_parser)
+        |> to_bin_output(:audio),
+        #
+        link(:demuxer)
+        |> via_out(Pad.ref(:video, 0))
+        |> to(:video_parser)
+        |> to_bin_output(:video)
```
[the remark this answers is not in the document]
shuntrho: It is possible, however for now I think we might just require linking both audio and video. If only one of them is needed, we can simply connect fake sink. Should this funcionality be needed, we may add it in the future.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,267 @@
...
+  defp get_flv_body(
+         %Membrane.RTMP.Header{
+           timestamp: timestamp,
+           body_size: data_size,
+           type_id: type_id,
+           stream_id: stream_id
+         },
+         payload
+       ) do
+    <<0::32, type_id::8, data_size::24, timestamp::24, 0::8, stream_id::24,
```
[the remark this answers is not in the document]
shuntrho: That's right, the value `0` should be sent only before the first packet.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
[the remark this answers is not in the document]
shuntrho: Okay, in fact what we are getting here is in fact a single tag, not a whole body.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -0,0 +1,267 @@
...
+    |> send_rtmp_payload(state.socket, chunk_size, chunk_stream_id: 3)
+
+    Responses.on_bw_done()
+    |> send_rtmp_payload(state.socket, chunk_size, chunk_stream_id: 3)
+
+    {:cont, %{state | message_parser: message_parser}}
+  end
+
+  defp do_handle_client_message(
+         %Messages.ReleaseStream{tx_id: tx_id, stream_key: _stream_key} = msg,
```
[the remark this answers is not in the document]
shuntrho: I added some descriptions for messages not described in RTMP documentation.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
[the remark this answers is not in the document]
daniel-jodlos: Ok, that makes perfect sense. A small suggestion though, I think you can use a function from FLV serializer to make this clause a little easier to ready for uninitiated.
https://github.com/membraneframework/membrane_flv_plugin/blob/7d3c32c2a401c80de4f3555e65250e0321bce85d/lib/membrane_flv_plugin/serializer.ex#L11

daniel-jodlos (approved):

PR #31, opened 2022-10-17
Fix memory leak

dmorn:
- Resolve memory leak with av_interleaved_write_frame
- Add failing test
- Implement packet interleave logic in pure Elixir
- Fix demands and reformat code
- Reorder functions
- Use ceil because buffer pts contain something that is not an integer
- Use av_packet_from_data in order to have a reference counted buffer


shuntrho:
Why exactly did you write your own interleaving?
From my understanding the changes is the native code fixed the memory leak, so why are there any changes in elixir code?
I find it highly unlikely that the `av_interleaved_write_frame` is the issue here.

dmorn:
Hi @shuntrho! What you're talking about is the `fix/memory-leak-with-native-interleaving`. If you send that stream through a "slow" channel (i.e. non a local network), the receiver won't get the payload contents. In our case was AWS Medialive reporting no audio/video detected. So our change does not play well with that function. Further, this plugin was already doing some interleaving with the demand. As soon as [avformat](https://github.com/libav/libav/blob/c4642788e83b0858bca449f9b6e71ddb015dfa5d/libavformat/avformat.h) tells to prefer `av_write_frame` when interleaving is already performed AND it is easier to manage the memory with that function (and understand what is going on), we went for it!

dmorn:
> If you send that stream through a "slow" channel (i.e. non a local network), the receiver won't get the payload contents

This behaviour is kind of difficult to catch: if you send everything to a local ffmpeg RTMP server everything is fine. If it takes "some time" to send the packets out in the network instead, the payloads get freed before going out apparently!

mix.exs
```diff
@@ -41,11 +41,13 @@ defmodule Membrane.RTMP.Mixfile do
       {:membrane_core, "~> 0.10.0"},
       {:unifex, "~> 1.0"},
       {:membrane_h264_ffmpeg_plugin, "~> 0.21.1"},
-      {:membrane_aac_plugin, "~> 0.12.1"},
+      {:membrane_aac_plugin, github: "membraneframework/membrane_aac_plugin", override: true},
```
philipgiuliani, line 44: Note that we have switched to the aac plugin, so the timestamps of our sample .msr file are not being overridden by the aac parser (see last commit). We are still waiting for the next release on hex.

shuntrho: Ok, I released the latest changes on hex.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -12,16 +12,17 @@ defmodule Membrane.RTMP.Sink do
...
 
   @supported_protocols ["rtmp://", "rtmps://"]
   @connection_attempt_interval 500
   @default_state %{
     attempts: 0,
     native: nil,
-    buffered_frame: nil,
+    # Keys here are the pad names.
+    frame_buffer: %{audio: nil, video: nil},
     ready: false,
```
shuntrho, line 24: I know you didn't change this code, but you can change this by the way.
```suggestion
    ready?: false,
```

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -12,16 +12,17 @@ defmodule Membrane.RTMP.Sink do
...
-    current_timestamps: %{}
+    forward_mode: false
```
shuntrho, line 25: I am not so sure about the name, maybe use `end_of_stream?` or add a comment.
```suggestion
    forward_mode?: false
```

shuntrho (changes requested):
In general, the change looks good.

My main comment is that we should raise errors if we do not intend to handle them.
Other than that, I made some minor comments and nitpicks.

mix.exs
```diff
@@ -41,11 +41,13 @@ defmodule Membrane.RTMP.Mixfile do
...
       {:membrane_mp4_plugin, "~> 0.16.0"},
+      {:membrane_file_plugin, "~> 0.12.0"},
```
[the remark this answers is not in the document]
philipgiuliani: We have initially added it to test only, but the `membrane_mp4_plugin` has a dependency on it, so I couldn't add it just for the `test` environment in this case: https://github.com/membraneframework/membrane_mp4_plugin/blob/master/mix.exs#L46

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -187,22 +197,78 @@ defmodule Membrane.RTMP.Sink do
...
+
+  defp fill_frame_buffer(state, pad, buffer) do
+    if get_in(state, [:frame_buffer, pad]) != nil do
+      {:error, "attempted to overwrite frame buffer on pad #{inspect(pad)}"}
+    else
+      {:ok, put_in(state, [:frame_buffer, pad], buffer)}
+    end
+  end
+
+  defp write_frame_interleaved(state = %{frame_buffer: fb = %{audio: audio, video: video}}) do
```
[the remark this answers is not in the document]
dmorn: both done.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
dmorn: I added a comment!

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -171,7 +171,6 @@ UNIFEX_TERM write_video_frame(UnifexEnv *env, State *state,
 
 end:
   av_packet_unref(packet);
-  av_packet_free(&packet);
```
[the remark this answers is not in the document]
dmorn: visualization(4)
No leakage and video/audio payloads are present in final output!

shuntrho (approved):

shuntrho (approved):
Looks great! 🥇 
Just making sure that these are all the changes in the PR and I'm merging it!

daniel-jodlos (approved):

PR #32, opened 2022-10-18
Improved validator

shuntrho:


lib/membrane_rtmp_plugin/rtmp/source/message_validator.ex
```diff
@@ -14,30 +14,30 @@ defmodule Membrane.RTMP.MessageValidator do
...
   """
-  @callback validate_release_stream(Messages.ReleaseStream.t()) :: :ok | {:error, any()}
+  @callback validate_release_stream(Messages.ReleaseStream.t()) ::
+              {:ok | :error, message :: any()}
 
   @doc """
   Validates the `t:Membrane.RTMP.Messages.Publish.t/0` message.
   """
-  @callback validate_publish(Messages.Publish.t()) :: :ok | {:error, any()}
+  @callback validate_publish(Messages.Publish.t()) :: {:ok | :error, message :: any()}
```
Qizot, line 37: II would keep `{:ok, result :: any()} | {:error, reason :: any()}`, it is a little bit more readable than trying to pack it into a single tuple with `|`.

Qizot: And you can probably pack that into `validation_result_t`.

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -244,6 +247,16 @@ defmodule Membrane.RTMP.MessageHandler do
     :gen_tcp.send(socket, [header | payload])
   end
 
+  defp get_validation_action(state, result) do
+    notification =
+      case result do
+        {:ok, msg} -> {:notify, {:stream_validation_success, msg}}
+        {:error, msg} -> {:notify, {:stream_validation_error, msg}}
```
Qizot, line 254: `reason` instead of `msg` would be more readable.

Qizot: And how about naming it `validation_action` instead of `get`? You are in fact not accessing but creating it so the `get` looks quite weird.

Qizot (changes requested):

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
shuntrho: okay, I just prefer to have some verb in the function name

Qizot (approved):

PR #39, opened 2023-01-04
Send notification on socket closed when the stream hasn't started

Qizot:
The following change ensures that the element hosting the RTMP Source will get notified when a socket closes without previous `start_of_stream` event. Previously the RTMP source would send `end_of_stream` event on socket close but it was ignored if the media really never flew through.

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/source.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/source.ex
@@ -54,6 +54,13 @@ defmodule Membrane.RTMP.Source do
   @type stream_validation_failed_t() ::
           {:stream_validation_failed, validation_stage_t(), reason :: any()}
 
+  @typedoc """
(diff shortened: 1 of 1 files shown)


shuntrho:
I am not sure about this solution. If some data is received by the Source but not enough to constitute to a full message, then afaik `start_of_stream` won't be received on the output pads. However, in such case we would send the notification anyway.

Qizot:
As far as I know the `start_of_stream` is emitted before the first buffer (but is followed by the buffer). In this case if we have a closed socket that won't give us any more data so we won't produce any sensible RTMP packet, we won't ever produce `start_of_stream` so the last resort is to inform the parent that we are done.

Other solution suggested on DMs by @bblaszkow06 was to send a `start_of_stream` event and immediately after it `end_of_stream` but for me this does not seem right as in fact we never start nor stop the stream in that case.

shuntrho (approved):
LGTM 💯

PR #40, opened 2023-01-09
Add support for RTMPS

Qizot:


Qizot:
I've added tests for RTMPS but the problem is how to keep the certificates to use them for testing? I would go for creating them in CI and then passing their paths via environmental variables but I have no idea how I can extend already existing CricleCI jobs to generate certs there and put them in certain directory so it is available for the next job.

lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -143,9 +190,16 @@ defmodule Membrane.RTMP.Source do
   end
 
   def handle_info(:start_receiving, _ctx, %{socket_retries: retries} = state) do
-    case :gen_tcp.controlling_process(state.socket, state.receiver_pid) do
+    module = if state.use_ssl?, do: :ssl, else: :gen_tcp
+
+    case module.controlling_process(state.socket, state.receiver_pid) do
       :ok ->
-        :ok = :inet.setopts(state.socket, active: :once)
+        if state.use_ssl? do
```
shuntrho, line 197: Why not use `module.setops(...)` here?

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -293,4 +297,8 @@ defmodule Membrane.RTMP.MessageHandler do
         parse_packet_messages(<<>>, message_parser, [{nil, step} | messages])
     end
   end
+
+  @compile {:inline, socket_module: 1}
```
shuntrho, line 301: I wonder if we really need this little bit of performance here, I leave this decision up to you

test/test_helper.exs
```diff
@@ -1 +1 @@
-ExUnit.start(capture_log: true)
+ExUnit.start(capture_log: true, exclude: [:rtmps])
```
shuntrho, line 1: Why do we exclude `rtmps` here?

shuntrho (changes requested):

test/test_helper.exs
Qizot: Read the comment I left above.

lib/membrane_rtmp_plugin/rtmp/source/source.ex
Qizot: because `:ssl` module doesn't work with that

shuntrho (approved):

PR #42, opened 2023-03-31
Add timestamp delta persistence between packet headers

Qizot:
Accordingly to the RTMP spec we were missing reusing a timestamp delta for `TYPE 3` packets which resulted
in timestamp duplication between following packets which with audio and video resulted in a timestamp drift that could never be recovered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The package can be installed by adding `membrane_rtmp_plugin` to your list of de
 ```elixir
 def deps do
   [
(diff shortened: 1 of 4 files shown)


bblaszkow06 (approved):
Wow, nice find!

shuntrho (approved):
LGTM 🥇

dmorn:
I was debugging this very same thing right now 😆 Thanks!

PR #43, opened 2023-04-04
Add support for extended timestamps

Qizot:
This PR aims to add support for extended timestamps.

An extended timestamp is a timestamp that is sent via RTMP once the 24 bits of header's timestamp gets overflowed. This
happens approximately after 4h 40 minutes (number of milliseconds that can be stored on 24 bits). It can affect all existing header types and can affect the following packets.

Besides adding extended timestamps handling for RTMP headers the full timestamp gets passed when generating FLV tag (which again was only using 24 bits of the timestamp).


test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -232,6 +263,17 @@ defmodule Membrane.RTMP.SourceBin.IntegrationTest do
...
+        if ts_offset = Keyword.get(opts, :ts_offset) do
+          add_file_option(command, %FFmpex.Option{
+            name: "-output_ts_offset",
+            argument: ts_offset,
+            contexts: [:output]
+          })
+        else
+          command
+        end
+      end)
```
shuntrho, line 276: I kind of don't like the way this command is conditionally added, but I don't have a particular idea how to make it simpler either.

lib/membrane_rtmp_plugin/rtmp/source/message.ex
```diff
@@ -71,18 +71,22 @@ defmodule Membrane.RTMP.Message do
...
+    case payload do
+      <<chunk::binary-size(chunk_size), rest::binary>> ->
+        acc = [<<0b11::2, chunk_stream_id::6>>, chunk | acc]
+
+        do_chunk_payload(rest, chunk_stream_id, chunk_size, acc)
+
+      payload ->
+        [payload | acc]
+        |> Enum.reverse()
+    end
```
shuntrho, line 89: Nice refactor

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -85,6 +85,37 @@ defmodule Membrane.RTMP.SourceBin.IntegrationTest do
     assert :ok = Task.await(ffmpeg_task)
   end
 
+  # maximum timestamp in milliseconds
+  @extended_timestamp_tag 0xFFFFFF
+  test "SourceBin gracefully handles timestamp overflow" do
+    # offset half a second second in the past
```
shuntrho, line 91:
```suggestion
    # offset half a second in the past
```

shuntrho (approved):
I commented a nitpick, but other than that looks good to me 👍🏽

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
Qizot: I will try to change it

Issue #45, opened 2023-05-02
RTMP Source does not work with AWS MediaLive


philipgiuliani:
When receiving an AWS MediaLive RTMP stream, the library crashes with the following exception:

```
** (ArgumentError) the following keys must also be given when building struct Membrane.RTMP.Messages.SetDataFrame: [:duration, :file_size, :encoder]
--
(membrane_rtmp_plugin 0.11.1) Membrane.RTMP.Messages.SetDataFrame.__struct__/1
(elixir 1.14.2) lib/kernel.ex:2336: Kernel.struct!/2
(membrane_rtmp_plugin 0.11.1) lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex:162: Membrane.RTMP.MessageParser.read_frame/3
(membrane_rtmp_plugin 0.11.1) lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex:70: Membrane.RTMP.MessageParser.handle_packet/2
(membrane_rtmp_plugin 0.11.1) lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex:285: Membrane.RTMP.MessageHandler.parse_packet_messages/3
(pepe 0.1.0) lib/pepe/rtmp/source.ex:69: Pepe.RTMP.Source.handle_info/3
(membrane_core 0.11.3) lib/membrane/core/callback_handler.ex:138: Membrane.Core.CallbackHandler.exec_callback/4
(membrane_core 0.11.3) lib/membrane/core/callback_handler.ex:62: Membrane.Core.CallbackHandler.exec_and_handle_callback/5
Last message: {:tcp, #Port<0.20>, <<4, 0, 0, 0, 0, 0, 184, 18, 0, 0, 0, 0, 2, 0, 13, 64, 115, 101, 116, 68, 97, 116, 97, 70, 114, 97, 109, 101, 2, 0, 10, 111, 110, 77, 101, 116, 97, 68, 97, 116, 97, 3, 0, 12, 97, 117, 100, ...>>}
[...]

PR #46, opened 2023-05-03
Avoid enforcing duration, file_size and encoder presence

dmorn:
This fixes https://github.com/membraneframework/membrane_rtmp_plugin/issues/45 and makes it possible to ingest an RTMP stream created by AWS Medialive in a Membrane pipeline. Notice also that while ffmpeg does deliver those keys/value pairs in its AMF payload they are set to 0.0 except for the encoder. We don't foresee any unwelcome side effect.


roznawsk (approved):
Alright, looks good and doesn't seem likely to break anything.

PR #49, opened 2023-05-16
Redesign MessageValidator as a protocol

### LINKED ISSUE #47: MessageValidator API does not support stateful validators
Given the API provided in https://github.com/membraneframework/membrane_rtmp_plugin/blob/master/lib/membrane_rtmp_plugin/rtmp/source/message_validator.ex, it is not possible to design a stream validator that has a state, e.g. a pre-shared RTMP stream key that the server is going to accept. I would propose to redesign it as a Protocol!

dmorn:
Closes #47


dmorn:
Fixing linter errors...

dmorn:
OK, dialyzer is not passing because of the default implemention of the procotol (which we want) that never returns the {:error, _} case.. if I remove that it works. This is plain wrong, I used this code :D

DominikWolek:
I mean, if `{:error, _}` is never returned after your redesign, you can change `validation_result_t` to be simply equal to `message` and refactor `message_handler` accordingly.

dmorn:
Hi @DominikWolek... actually, implementers of the protocol might return the {:error, _} tuple as before to indicate that there was a validation error. It just that that such a Protocol consolidation does not exist in this codebase so it looks like (from dialyzer's point of view) that that branch can never be reached. I'll have a look at it later

dmorn:
@DominikWolek this implementation does not provide a default validator, which means that every user of the plugin need to implement their own. Do you want me to write a default one that allows every stream to be accepted or is this a OK for now?

roznawsk (changes requested):
Please update the spec for the MessageValidator in the SourceBin:
https://github.com/membraneframework/membrane_rtmp_plugin/blob/da72b94efe051ab478ce00b54e0adeb91039b661/lib/membrane_rtmp_plugin/rtmp/source/bin.ex#L53 ,
especially the default value - this should be a default validator implementation,
which accepts all streams.

While you are modifying the validator, you might consider adding a state to the stream validation process.
Now we are returning only `{:ok, result}` or `{:error, reason}` from the `validate_<message type>` callbacks. We could be returning the modified validator struct together with the result, e.g. `{:ok, result, validator}` or `{:error, reason, validator}`.
This would allow for fetching data from messages and using it when validating the messages that come next.

This however is a separate topic, so if you decide to implement it
you may do it in a separate PR.

lib/membrane_rtmp_plugin/rtmp/source/message_validator.ex
```diff
@@ -1,44 +1,30 @@
-defmodule Membrane.RTMP.MessageValidator do
+defprotocol Membrane.RTMP.MessageValidator do
```
Qizot, line 1: Is there any benefit to use the protocol instead of a simple behaviour? Using a protocol only adds some boilerplate without any other benefit (but I'm opened to conversation).

Qizot: Also with a behaviour we are able to provide default implementations for certain callbacks, with protocols it is no longer possible.

Qizot (changes requested):
Overall, I don't think that I'm a fan of the solution. We could easily pass the struct and operate on behaviours instead (so adding another param to callbacks just like with protocols).

dmorn:
@Qizot and @roznawsk, thanks for joining! Let's pause and ponder for a moment. What should the validator do? How is it used in a standard usecase? For us, the validator checks the streamkey, and it usually does so by using a pre-shared secret , that's the reason why we needed some state in the first place, hence the need for the PR. What did you expect users to do with the validator?

Why a protocol? If you need a state in the behaviour, you need the usual "init" callback or something to put the implementer's state in the loop. I personally prefer how a Protocol defines such situation as the state is already part of the game, this is just a matter of personal taste though.

About returning the state of the validator instead of plain `{:ok, _} | :error` tuples: as soon as I do not consider that a normal usecase, I thought that it would be cleaner to just keep those return values -- if users want to implement something like a state-machine to decide whether to accept the stream or not, they can always start an Agent and use it to preserve the state, putting the pid in the Protocol implementer's struct.

Let me know what you prefer, I do not care, its your plugin after-all. I just want this pre-shared secret somewhere and drop the fork at some point :D Tell me and I'll implement it.

test/support/test_validator.ex
```diff
@@ -0,0 +1,32 @@
...
+      }) do
+    if stream_key == accepted_key do
+      {:ok, "correct stream key"}
+    else
+      {:error, "wrong stream key"}
+    end
+  end
+
+  @impl true
+  def validate_release_stream(_impl, _message), do: {:ok, "release stream success"}
```
Qizot, line 28: nitpick: can you do the `defdelegate` here so other people who would like to get inspired by the tests can look it up?

Qizot (approved):
One last nitpick and otherwise LGTM! (Also please bump the version in both `mix.exs` and readme, this time minor one, not patch).

test/support/test_validator.ex
dmorn: As far as I know, you cannot delegate in this case as the MessageValidator.Default implements those functions in its protocol implementation only, they do not appear as functions of the module... 🙈

dmorn: version bumped

dmorn:
This is weird, tests are not passing now that I merged with master, they do pass in master though 🤔 Maybe I broke something while merging, I'll have a look at it tomorrow.

test/support/test_validator.ex
Qizot: Please bump to `0.12.0` instead since we are changing API.

roznawsk (approved):

Qizot:
Good job again 😄

PR #50, opened 2023-05-19
Draft: Fix crash when the rtmp stream starts after the timestamp overflow

philipgiuliani:


lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex
```diff
@@ -175,19 +180,23 @@ defmodule Membrane.RTMP.MessageParser do
...
     if byte_size(body) <= chunk_size do
       body
     else
-      do_combine_body_chunks(body, chunk_size, [])
+      do_combine_body_chunks(body, chunk_size, [], header)
     end
   end
 
-  defp do_combine_body_chunks(body, chunk_size, acc) do
+  defp do_combine_body_chunks(body, chunk_size, acc, header) do
```
dmorn, line 191: We see a refactoring opportunity in here as there is code duplication with the deserialise function. Maybe a reduction till all bytes required by the header have been parsed?

test/membrane_rtmp_plugin/rtmp_source_bin_test.exs
```diff
@@ -116,6 +116,35 @@ defmodule Membrane.RTMP.SourceBin.IntegrationTest do
     assert :ok = Task.await(ffmpeg_task)
   end
 
+  test "SourceBin gracefully handles start when starting after timestamp overflow" do
+    # offset half a second in the future
```
Qizot, line 120: nitpick: I think it would be better to start some more time in the future to avoid any potential accidental rounding errors done in the test when passing the offset around

lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex
```diff
@@ -149,14 +149,19 @@ defmodule Membrane.RTMP.MessageParser do
...
           else
             body_size
           end
 
         case rest do
           <<body::binary-size(chunked_body_size), rest::binary>> ->
-            combined_body = combine_body_chunks(body, chunk_size)
+            combined_body = combine_body_chunks(body, chunk_size, header)
+
+            if byte_size(combined_body) != body_size do
```
philipgiuliani, line 162: This guard was very helpful to make sure that we passed the correct body to the deserialize_message function. Not sure if we should keep it or not. It currently makes credo fail. Let me know if I should remove it or extract the check into a private assert function.

lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex
Qizot: I would leave it for now, and if you really want it could be a matter of a separate PR.

Qizot (changes requested):
When you address all the comments I think we are good to go :)

lib/membrane_rtmp_plugin/rtmp/source/message_parser.ex
```diff
@@ -175,19 +180,23 @@ defmodule Membrane.RTMP.MessageParser do
...
-      do_combine_body_chunks(body, chunk_size, [])
+      do_combine_body_chunks(body, chunk_size, [], header)
     end
   end
 
-  defp do_combine_body_chunks(body, chunk_size, acc) do
+  defp do_combine_body_chunks(body, chunk_size, acc, header) do
     case body do
+      <<body::binary-size(chunk_size), 0b11::2, _chunk_stream_id::6, timestamp::32, rest::binary>>
+      when header.extended_timestamp? and timestamp == header.timestamp ->
```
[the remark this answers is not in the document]
philipgiuliani: I am not sure how to do this, because the header.timestamp is an integer and not a binary. If matching a short integer it may parse less than 32 bits from the binary and will never match 🤔

I also think its nicer, otherwise I have to extract the timestamp from the header into an own variable.

philipgiuliani:
Version has been bumped

PR #53, opened 2023-05-25
Send end_of_stream on deleteStream message

### LINKED ISSUE #52: end_of_stream is sent when the TCP connection is dropped
Consider the case where a pipeline is reading from RTMP and outputting an m3u8 playlist. When the `end_of_stream` message is delivered, the sink appends the #EXT_X_ENDLIST tag to the playlist marking it as complete.

With the current logic, this event happens not at the actual end of the input stream but also when the TCP connection is dropped, which may happen for unrelated reasons such as a network fault.

dmorn:
This PR adds end_of_stream action in case of deleteStream and remove end_of_stream action
when the TCP connection is closed. Closes #52

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/source.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/source.ex
@@ -222,10 +222,10 @@ defmodule Membrane.RTMP.Source do
 
   @impl true
   def handle_info({:socket_closed, _socket}, ctx, state) do
-    if ctx.pads.output.start_of_stream? do
-      {[end_of_stream: :output], state}
-    else
-      {[notify_parent: :unexpected_socket_closed], state}
+    cond do
+      ctx.pads.output.end_of_stream? -> {[], state}
+      ctx.pads.output.start_of_stream? -> {[end_of_stream: :output], state}
+      true -> {[notify_parent: :unexpected_socket_closed], state}
     end
   end
 
--- a/lib/membrane_rtmp_plugin/rtmp/source/source.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/source.ex
@@ -223,8 +223,7 @@ defmodule Membrane.RTMP.Source do
   @impl true
   def handle_info({:socket_closed, _socket}, ctx, state) do
     if ctx.pads.output.start_of_stream? do
-      # EOS is handled by the message handler.
-      {[], state}
+      {[end_of_stream: :output], state}
     else
       {[notify_parent: :unexpected_socket_closed], state}
     end
--- a/lib/membrane_rtmp_plugin/rtmp/source/source.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/source.ex
@@ -223,7 +223,8 @@ defmodule Membrane.RTMP.Source do
   @impl true
   def handle_info({:socket_closed, _socket}, ctx, state) do
     if ctx.pads.output.start_of_stream? do
-      {[end_of_stream: :output], state}
+      # EOS is handled by the message handler.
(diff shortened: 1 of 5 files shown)


lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -223,7 +223,8 @@ defmodule Membrane.RTMP.Source do
   @impl true
   def handle_info({:socket_closed, _socket}, ctx, state) do
     if ctx.pads.output.start_of_stream? do
-      {[end_of_stream: :output], state}
+      # EOS is handled by the message handler.
+      {[], state}
```
Qizot, line 227: Remember that a socket may get closed unexpectedly without sending the `deleteStream`. Add a check here that if the `end_of_stream?` has not been sent then send one.

Qizot (changes requested):

lib/membrane_rtmp_plugin/rtmp/source/source.ex
dmorn: @Qizot that's exactly what we want to avoid! The socket might get closed unexpectedly yes, but that definitely does not mean that the RTMP stream is finished!

Qizot: Socket gets passed to the element just once so there is no way for continuing the stream, that is what `end_of_stream` is for (in membrane's philosophy). 

If you really want to act on `deleteStream` we maybe add that to the stream validator as another callback.

dmorn: In our codebase we're not actually using this source component but just the MessageValidator, we're raising the when connection get's closed so that the pipeline restarts, re-opens the TCP connection and continues where it left! I can share the whole module if you want.

The point is: you actually never want to signal an end_of_stream if its not, and a TCP connection closed isn't. It has major consequences that sometimes are irrecoverable, the one in the issue description is an example.

Qizot: I see it that way: most users don't care if the stream ends with `deleteStream` or with a socket being closed, majority of the time closing a socket will be preceded with the `deleteStream` (but for instance OBS sometimes doesn't care if you stop manually). If you have software such as OBS that crashes it won't send the `deleteStream` but once you open it once again it automatically sends `releaseStream`.

I guess you already have a workaround but to have this smart retry you could have a parent bin that wraps the RTMP source element + a smart filter.

Parent bin will receive notifications from RTMP such as mentioned `deleteStream` and in that case, it will communicate with the smart filter to send `end_of_stream` out of its output pad. If it receives socket closed instead it will try to restart the RTMP and in this case, the smart filter will ignore the `end_of_stream` from the RTMP element so it doesn't propagate further.

dmorn: OK, I really don't care actually. Can we make the MessageHandler send the end_of_stream when a deleteStream is received though? 🙌 I can revert the change in the source, or at least make it send the end_of_stream if it was not sent already.

Qizot: If you mean sending end of stream on `deleteStream` and also on socket close when `deleteStream` didn't do that already then go ahead.

dmorn: Done, version bumped!

dmorn:
After this change is merged, we can finally drop our fork 🙌

dmorn:
Done!

Qizot (approved):
LGTM!

PR #54, opened 2023-06-14
Sink single track

roznawsk:
Changed RTMP Sink input pads availability to `:on_request`, thus allowing for sending only a single track (video or audio).


c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -7,10 +7,15 @@ void handle_init_state(State *state);
 
 void handle_destroy_state(UnifexEnv *env, State *state);
 
-UNIFEX_TERM create(UnifexEnv *env, char *rtmp_url) {
+int is_ready(State *state);
```
Qizot, line 10:
```suggestion
static int is_ready(State *state);
```

examples/sink_video.exs
```diff
@@ -0,0 +1,63 @@
...
+      |> child(:video_parser, %Membrane.H264.FFmpeg.Parser{
+        framerate: {25, 1},
+        alignment: :au,
+        attach_nalus?: true,
+        skip_until_keyframe?: true
+      })
+      |> child(:video_realtimer, Membrane.Realtimer)
+      |> child(:video_payloader, Membrane.MP4.Payloader.H264)
+      |> via_in(Pad.ref(:video, 0))
+      |> get_child(:rtmp_sink)
```
Qizot, line 34: You can probably put the `child(:rmtp_sink...` here straight away.

Qizot (changes requested):
Please bump the versions both in README and in mix.exs. Besides that after those small suggestions I think we are fine to merge.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -7,10 +7,15 @@ void handle_init_state(State *state);
...
+
+UNIFEX_TERM create(UnifexEnv *env, char *rtmp_url, int audio_present, int video_present) {
```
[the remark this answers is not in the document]
roznawsk: The `create` function parameters have to be of type `int` since unifex maps bool to int, however I will change it inside the `State` struct.

c_src/membrane_rtmp_plugin/sink/rtmp_sink.c
```diff
@@ -7,10 +7,15 @@ void handle_init_state(State *state);
 
 void handle_destroy_state(UnifexEnv *env, State *state);
```
Qizot, line 8: I know that it has been there before, but we probably don't need the function forward declaration as it is already present in auto-generated code, am I right?

examples/sink_video.exs
```diff
@@ -0,0 +1,62 @@
...
+  @samples_url "https://raw.githubusercontent.com/membraneframework/static/gh-pages/samples/big-buck-bunny/bun33s_480x270.h264"
+
+  @impl true
+  def handle_init(_ctx, destination: destination) do
+    structure = [
+      child(:video_source, %Membrane.Hackney.Source{
+        location: @video_url,
+        hackney_opts: [follow_redirect: true]
+      })
+      |> child(:video_parser, %Membrane.H264.FFmpeg.Parser{
```
Qizot, line 24: Don't we want to use elixir parser in most places at the moment?

examples/sink_video.exs
```diff
@@ -0,0 +1,62 @@
...
+  @impl true
+  def handle_element_end_of_stream(_child, _pad, _ctx, state) do
+    {[], state}
+  end
+end
+
+destination = System.get_env("RTMP_URL", "rtmp://localhost:1935")
+
+# Initialize the pipeline and start it
+{:ok, _supervisor, pipeline} = Example.start_link(destination: destination)
```
Qizot, line 54: Not related to this PR, but it would probably be cool to have the functionality to await pipeline termination in core's API.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -97,37 +126,52 @@ defmodule Membrane.RTMP.SinkTest do
...
-      ],
+      structure:
+        [
+          child(:rtmp_sink, %Membrane.RTMP.Sink{
+            rtmp_url: rtmp_url,
+            tracks: tracks,
+            max_attempts: 5
+          })
+        ] ++
+          if :audio in tracks do
```
Qizot, line 141: nitpick: just for readability we can probably extract this (+ the video below) to a helper function with a guard `when :video in tracks/ when :audio in tracks`

Qizot (approved):
Nothing blocking, nice job!

examples/sink_video.exs
varsill: I am just worried how crucial is the `skip_until_keyframe?` option in that case

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -45,26 +50,50 @@ defmodule Membrane.RTMP.SinkTest do
...
     # so it may retry a few times before succeeding
     assert_pipeline_play(sink_pipeline_pid, 5000)
 
-    assert_start_of_stream(sink_pipeline_pid, :rtmp_sink, :video, 5_000)
-    assert_start_of_stream(sink_pipeline_pid, :rtmp_sink, :audio, 5_000)
+    if :audio in tracks,
+      do: assert_start_of_stream(sink_pipeline_pid, :rtmp_sink, Pad.ref(:audio, 0), 5_000)
+
+    if :video in tracks,
+      do: assert_start_of_stream(sink_pipeline_pid, :rtmp_sink, Pad.ref(:video, 0), 5_000)
```
varsill, line 83: How about: 
```suggestion
Enum.each(tracks, &assert_start_of_stream(sink_pipeline_pid, :rtmp_sink, Pad.ref(&1, 0), 5_000))
```

test/membrane_rtmp_plugin/rtmp_sink_test.exs
roznawsk: Unfortunately, I cannot use guards, because the `tracks` are unknown at compile time. I will move it to a separate function though.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -45,26 +50,50 @@ defmodule Membrane.RTMP.SinkTest do
...
+      do: assert_end_of_stream(sink_pipeline_pid, :rtmp_sink, Pad.ref(:video, 0), 5_000)
 
     :ok = Pipeline.terminate(sink_pipeline_pid, blocking?: true)
     # RTMP server should terminate when the connection is closed
     assert :ok = Task.await(rtmp_server)
 
-    assert File.exists?(flv_output_file)
-    assert File.stat!(flv_output_file).size == File.stat!(@reference_flv_path).size
+    assert File.exists?(flv_output)
+    assert File.stat!(flv_output).size == File.stat!(fixture).size
```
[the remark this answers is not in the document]
roznawsk: Hmm, for one it doesn't work. Here's the diff between the fixture and the output for a single video track test (btw, that's the only test which fails): 
image

examples/sink_video.exs
roznawsk: I guess that it would work fine when using files as input, since there is probably keyframe at the beginning. Problems could arise when connecting in a middle of a stream.

test/membrane_rtmp_plugin/rtmp_sink_test.exs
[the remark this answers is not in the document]
varsill: Since the single video track test is newly added test, perhaps it would be ok to make a "snapshot" test and provide the output of our pipeline as the reference result? Surely, we should also manually confirm the the output .flv of our pipeline is correct. Now it won't check anything, but it will bring benefits with further changes so that to make sure the result is not changing.

examples/sink_video.exs
varsill: Indeed, but we cannot assume that the stream we receive in RTMP Sink is a stream from a file (despite the fact that in all the tests we are using only streams from the file)

test/membrane_rtmp_plugin/rtmp_sink_test.exs
```diff
@@ -60,6 +60,7 @@ defmodule Membrane.RTMP.SinkTest do
     test_stream_correctly_received(flv_output_file, @reference_flv_audio_path, [:audio])
   end
 
+  @tag :this
```
varsill, line 63:
```suggestion
```
probably a debugging leftover ;)

varsill (changes requested):

varsill (approved):
Great job, LGTM 🥇

PR #56, opened 2023-06-16
Close avio stream instead of just flushing pending buffers

dmorn:
Right now, if we stream to an RTMP server (e.g. ffmpeg) the connection will not be properly
closed on finalization. This is because we're only flushing the IO stream instead of flushing AND
closing.

This is how finalization works according to the [docs](https://github.com/FFmpeg/FFmpeg/blob/12a073f72cf146d16c7987cf8782519b69be5f56/libavformat/avformat.h#LL235C4-L238C28):
```
Once all the data has been written, the caller must call av_write_trailer()
to flush any buffered packets and finalize the output file, then close the IO
context (if any) and finally free the muxing context with
avformat_free_context().
```

There are two missing bits in the current implementation:
- [x] close the IO context
- [x] free it

I'm testing the free, for now I'm opening the PR for early feedback. I'm testing it and ffmpeg
now exists normally.


dmorn:
Getting a segfault on the tests. I'll setup Rambo locally (awful experience with m2 macs). Update coming soon.

dmorn:
Ah, the problem is that I do not think this bit of code does what you would expect
```
  # for Mac M1 it is necessary to include rambo compiler (used by :ffmpex)
  if Mix.env() == :test do
    def maybe_add_rambo(), do: [:rambo]
  else
    def maybe_add_rambo(), do: []
  end
```
Mix.env() is set to :dev if you run `mix test` 🤷‍♂️

dmorn:
`MIX_ENV=test mix test` once then the compiler will be available, still you may consider checking this out.

dmorn:
Ahhhh I see now where the segfault comes from. There is
```
void handle_destroy_state(UnifexEnv *env, State *state) {
  UNIFEX_UNUSED(env);
  if (state->output_ctx &&
      !(state->output_ctx->oformat->flags & AVFMT_NOFILE)) {
    avio_closep(&state->output_ctx->pb);
  }
  if (state->output_ctx) {
    avformat_free_context(state->output_ctx);
  }
}
```
which is eventually closing it. Apparently in my case this function was not called, at least the `avio_close` part.

dmorn:
Note that on master I get
```
Finished in 13.8 seconds (13.8s async, 0.00s sync)
12 tests, 3 failures, 2 excluded
```
When I run the tests. The problem is the size comparison of two fixtures, 1 missing byte in each one of them (newline?)

Qizot:
To trigger the `handle_destroy_state` we can in `handle_end_of_stream` assign `nil` to the native reference. This should trigger the garbage collection quite fast resulting in all those closing function getting called.

Qizot:
@dmorn look up my previous comment if that approach fixes the problem.

Qizot:
The element should probably be used as-is so if we unassign the sink reference in end of stream to trigger garbage collection should be fine. Note that we need to close the structures in `handle_state_destroy` as the pipeline may crash and then you won't call `finalize_stream`, the question is if it should flush the connection there too.

dmorn:
>  the question is if it should flush the connection there too.
Maybe that's the only thing that should be changed! flush on finalize, release on `handle_state_destroy`

Qizot:
So we probably just need to assign to `nil` in `handle_end_of_stream` and the rest of the C code stays as it is.

Qizot:
Okay, let's do it that way. Add a flag to the state `closed`. Change the `finalize` to `flush_and_close` (that function will set the `closed` flag) that will be called in the `handle_end_of_stream`. In `handle_destroy_state` check if `closed` flag is set, if not then basically call the `flush_and_close` and be done with it.

This way we have an understandable API (good mental model when ending the stream) and we prevent any resource leaking once the native resource gets garbage collected). WDYT?

dmorn:
Sounds good to me @Qizot! I' ll update the PR soon

README.md
```diff
@@ -14,7 +14,7 @@ The package can be installed by adding `membrane_rtmp_plugin` to your list of de
 ```elixir
 def deps do
   [
-	  {:membrane_rtmp_plugin, "~> 0.14.0"}
+	  {:membrane_rtmp_plugin, "~> 0.14.1"}
```
Qizot, line 17: What do you think about bumping it to `0.15.0` and getting rid of `finalize_stream`?

Qizot (changes requested):

README.md
dmorn: I wouldn't break the public API because I think
- `finalize_stream` is OKish, I find it intuitive
- user will get the benefit of the new implementation w/o code changes
- I see this more as a bugfix rather than a feature update

Said that, we can get rid of that function if you prefer to!

Qizot: Ok, seems fair enough :)

Qizot (approved):
LGTM. Thanks for contributing so many fixes, greatly appreciated!

PR #57, opened 2023-06-22
Send to parent the RTMP connect message

enzoqtvf:
Hey team, this is not really a finished PR but something that I would like to have from this library.
In this PR I'm modifying the message_handler to send the RTMP.Message.Connect to the parent as a notification.
I found it useful in one of my project where I can target my RTMP server with urls like this `rtmp://localhost:<port>/?some_id=AAA` and I can get this ID into my pipeline.
If there is another way to do it I'm happy to explore it too :)
Off course if you like this idea we can add test and doc :)

Thanks !

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
@@ -76,7 +76,7 @@ defmodule Membrane.RTMP.MessageHandler do
     {:cont, %{state | message_parser: parser}}
   end
 
-  defp do_handle_client_message(%Messages.Connect{}, _header, state) do
+  defp do_handle_client_message(%Messages.Connect{} = connect, _header, state) do
     chunk_size = state.message_parser.chunk_size
 
     [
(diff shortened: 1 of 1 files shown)


Qizot:
The same key is sent in `Membrane.RTMP.Messages.Publish` `stream_key` field and that you can already handle with `MessageValidator`. Can you check if that suits you?

enzoqtvf:
Oh didn't know that, checking it ! and letting you know :) !

enzoqtvf:
Ah I see, so if I'm not wrong I can use a url like this: `rtmp://localhost:<port>/app/?some_id=AAA` and my stream_key will be `?some_id=AAA`which I can catch in my Validator, however from there I guess I would need to use some ETS table so my pipeline can find it, I don't see how I can use the Validator to send it back to the pipeline otherwise.

Qizot:
You can save the pipeline pid in your `MessageValidator` struct (you have to implement the validator's protocol) and in the callback implementation you can send a messagse to the pipeline and catch that with `handle_info`.

Qizot:
Closing it for now. If you still have problems getting it to work then I will reopen.

PR #58, opened 2023-06-24
fix: waiting other frames.

mohammedzeglam-pg:


Qizot:
Thanks for catching up the bug :) 

Since the sink can now send a single track only this `write_frame_interleaved` is no longer valid as we have to check if the tracks are present. I will create a different fix that will aim to do what you did.

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
```diff
@@ -253,7 +253,11 @@ defmodule Membrane.RTMP.Sink do
     end
   end
 
-  defp write_frame_interleaved(%{frame_buffer: %{audio: audio, video: video}} = state)
+  defp write_frame_interleaved(
+         %{
+           frame_buffer: %{{Membrane.Pad, :audio, 0} => audio, {Membrane.Pad, :video, 0} => video}
```
Qizot, line 258:
```suggestion
           frame_buffer: %{Pad.ref(:audio, 0) => audio, Pad.ref(:video, 0) => video}
```

Qizot (changes requested):
Thanks for catching up the bug :)

Can you bump the patch version as well in `mix.exs` and `README.md`?

lib/membrane_rtmp_plugin/rtmp/sink/sink.ex
mohammedzeglam-pg: Thanks for help me with this. and sorry for take your time to guide me.

Qizot: No need to be sorry, you are more than fine :)

Qizot (approved):

PR #63, opened 2023-07-18
Add support for `additionalMedia` message

Qizot:
This PR introduces support for `additionalMedia` AMF data message.

It is being used in OBS by Twitch to send a second audio track that can be used for VOD purposes (replacing live audio that can be copyrighted with a non-copyrighted one).

To introduce the change `AMF3` parser is needed (we only care about the byte array type, which is why many types are not implemented).

## TODO:
- [X] transform `AdditionalMedia` message to a membrane's buffer
- [X] add callback to the message's validator to support handling modified `@setDataFrame` that indicates additional media track (it doesn't have to container `onMetaData`)
- [X] discover how we can demux the FLV stream when it contains another track (usually it can only carry one, and it is not obvious how it can be differentiated  between separate audio tracks)

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
@@ -1,0 +1,216 @@
+defmodule Membrane.RTMP.AMF3.Parser do
+  @moduledoc false
+
+  import Bitwise
+
+  @doc """
+  Parses message from AMF3 format to elixir data types.
+  """
+  @spec parse_list(binary()) :: list()
+  def parse_list(binary) do
+    do_parse(binary, [])
+  end
+
+  @doc """
+  Parses message from AMF3 format to elixir data type.
+  """
+  @spec parse(binary()) :: {value :: term(), rest :: binary()}
+  def parse(binary) do
+    parse_value(binary)
+  end
+
+  defp do_parse(<<>>, acc), do: Enum.reverse(acc)
+
+  defp do_parse(payload, acc) do
+    {value, rest} = parse_value(payload)
+
+    do_parse(rest, [value | acc])
+  end
+
+  # undefined
+  defp parse_value(<<0x00, rest::binary>>) do
+    {:undefined, rest}
+  end
+
+  # null
+  defp parse_value(<<0x01, rest::binary>>) do
+    {nil, rest}
+  end
+
+  # false
+  defp parse_value(<<0x02, rest::binary>>) do
+    {false, rest}
+  end
+
+  # true
+  defp parse_value(<<0x03, rest::binary>>) do
+    {true, rest}
+  end
+
+  # integer
+  defp parse_value(<<0x04, rest::binary>>) do
+    parse_integer(rest)
+  end
+
+  # double
+  defp parse_value(<<0x05, double::float-size(64), rest::binary>>) do
+    {double, rest}
+  end
+
+  # string
+  defp parse_value(<<0x06, rest::binary>>) do
+    parse_string(rest)
+  end
+
+  # xml document
+  defp parse_value(<<0x07, rest::binary>>) do
+    {size, rest} = parse_integer(rest)
+
+    # if the bit is set then we are dealing with string literal
+    if (size &&& 0x01) == 1 do
+      size = size >>> 1
+
+      <<string::binary-size(size), rest::binary>> = rest
+
+      {{:xml_doc, string}, rest}
+    else
+      raise "Unsupported xml doc reference"
+    end
+  end
+
+  # date
+  defp parse_value(<<0x08, rest::binary>>) do
+    {number, rest} = parse_integer(rest)
+
+    if (number &&& 0x01) == 1 do
+      {{:date, :previous}, rest}
+    else
+      <<date::float-size(64), rest::binary>> = rest
+
+      {{:date, date}, rest}
+    end
+  end
+
+  # array
+  defp parse_value(<<0x09, rest::binary>>) do
+    {size, rest} = parse_integer(rest)
+
+    if (size &&& 0x01) == 1 do
+      dense_array_size = size >>> 1
+
+      {assoc_array, rest} = parse_assoc_array(rest, [])
+      {dense_array, rest} = parse_dense_array(rest, dense_array_size, [])
+
+      {assoc_array ++ dense_array, rest}
+    else
+      raise "Unsupported array reference"
+    end
+  end
+
+  # object
+  defp parse_value(<<0x0A, _rest::binary>>) do
+    raise "Unsupported object type"
+  end
+
+  # xml
+  defp parse_value(<<0x0B, rest::binary>>) do
+    {size, rest} = parse_integer(rest)
+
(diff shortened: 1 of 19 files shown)


mohammedzeglam-pg:
could i help with this?

Qizot:
@mohammedzeglam-pg sorry for now responding quickly enough (I didn't get the github notification).

This is a really bare draft PR that I hope to finish in the future as this is a non-trivial handling of Twitch specific stuff implemented in OBS.

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+
+    # if the bit is set then we are dealing with string literal
+    if (size &&& 0x01) == 1 do
+      size = size >>> 1
+
+      <<string::binary-size(size), rest::binary>> = rest
+
+      {{:xml_doc, string}, rest}
+    else
+      raise "Unsupported xml doc reference"
```
balins, line 77: Why do we support Date references and don't support XML ones? If I understand correctly, we'd just need to parse the remaining 28 bits as an integer and return `{{:xml_ref, ref_integer}, rest}` (for now we could omit ref table implementation, but technically we could implement it in the future by keeping referenced objects in the process dictionary)

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+    end
+  end
+
+  # date
+  defp parse_value(<<0x08, rest::binary>>) do
+    {number, rest} = parse_integer(rest)
+
+    if (number &&& 0x01) == 1 do
+      {{:date, :previous}, rest}
```
balins, line 86: If I understand correctly it's not **the** previous date, but the following 28 bits of `number` contain the reference to **a** previous date referenced by this one.

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      {{:date, date}, rest}
+    end
+  end
+
+  # array
+  defp parse_value(<<0x09, rest::binary>>) do
+    {size, rest} = parse_integer(rest)
+
+    if (size &&& 0x01) == 1 do
```
balins, line 98: at this point it seems like we could extract it to a helper function `parse_reference_marker` that returns `{:reference, reference, rest}` or `{:value, unparsed, rest}` and `case` on it (or raise on `:reference` if we don't want to support them and return just `{unparsed, rest}`)

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+      dense_array_size = size >>> 1
+
+      {assoc_array, rest} = parse_assoc_array(rest, [])
+      {dense_array, rest} = parse_dense_array(rest, dense_array_size, [])
+
+      {assoc_array ++ dense_array, rest}
+    else
+      raise "Unsupported array reference"
```
balins, line 106: Seems like we could return a reference here as well?
Just to be clear, I don't insist on implementing reference handling, but if we don't want support it, then I'd raise on `:date` reference too.

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+    end
+  end
+
+  # object
+  defp parse_value(<<0x0A, _rest::binary>>) do
+    raise "Unsupported object type"
```
balins, line 112: nit: it's not clear that you mean literally **the** `object` type 
```suggestion
    raise "Unsupported AMF3 type: object"
```

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  # xml
+  defp parse_value(<<0x0B, rest::binary>>) do
```
balins, line 116: same code again?

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+  # date
+  defp parse_value(<<0x08, rest::binary>>) do
+    {number, rest} = parse_integer(rest)
+
+    if (number &&& 0x01) == 1 do
+      {{:date, :previous}, rest}
+    else
+      <<date::float-size(64), rest::binary>> = rest
+
+      {{:date, date}, rest}
```
balins, line 90: btw it would be more consistent to parse it as datetime and return plain Elixir type instead of `{:date, date}` (we don't return `{:assoc_array, ...}`, `{:integer, ...}` etc)

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      <<string::binary-size(size), rest::binary>> = rest
+
+      {string, rest}
+    else
+      raise "Unsupported string reference"
+    end
+  end
+
+  defp parse_assoc_array(<<0x01, rest::binary>>, acc), do: {Enum.reverse(acc), rest}
```
balins, line 200: Correct me if I'm wrong, but shouldn't we terminate parsing assoc array on an empty string? So the check should include `<<0x01, 0::28>>`

lib/membrane_rtmp_plugin/rtmp/source/message_handler.ex
```diff
@@ -258,31 +278,58 @@ defmodule Membrane.RTMP.MessageHandler do
...
+    payload = get_flv_tag(header, 1, media.media)
+
+    action = {:buffer, {:output, %Buffer{payload: payload}}}
+
+    Map.update!(state, :actions, &[action | &1])
+  end
+
+  defp get_additional_media_actions(rtmp_header, media, state) do
+    header = %Membrane.RTMP.Header{rtmp_header | type_id: 8, body_size: byte_size(media.media)}
+    payload = get_flv_header() <> get_flv_tag(header, 1, media.media)
```
balins, line 298: `media.media` 😞 Let's consider `payload` instead as the field name

balins (changes requested):

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
Qizot: there are `xml_doc` and `xml` types if you mean that

lib/membrane_rtmp_plugin/rtmp/source/amf3/parser.ex
Qizot: As we checked, the empty UTF-8 is just a `0x01` marker

balins (approved):

PR #66, opened 2023-08-17
Use the new parser

mat-hek:


Qizot (approved):
LGTM but I would wait with merging until the mp4 plugin gets released

PR #67, opened 2023-08-31
Update the plugin to use new formats and replace payloaders with parsers

Noarkhh:
- Replace MP4.Payload with H264 stream format
- Replace MP4.Payloaders with H264.Parsers


examples/sink.exs
```diff
@@ -28,26 +29,25 @@ defmodule Example do
...
       |> child(:audio_parser, %Membrane.AAC.Parser{
-        in_encapsulation: :ADTS,
         out_encapsulation: :none
       })
       |> child(:audio_realtimer, Membrane.Realtimer)
       |> via_in(Pad.ref(:audio, 0))
       |> get_child(:rtmp_sink)
     ]
 
-    {[spec: structure, playback: :playing], %{streams_to_end: 2}}
```
Qizot, line 50: Has something changed in the core that `playback: :playing` is no longer required?

Qizot (approved):
Please post links to PRs that made it possible to get rid of `MP4` payloaders.

examples/sink.exs
varsill: Indeed, it has been removed in https://github.com/membraneframework/membrane_core/releases/tag/v0.12.1

varsill (approved):

examples/sink.exs
Qizot: Good to know :)

PR #75, opened 2023-10-26
Use precompiled deps

DominikWolek:
- Change `bundlex.exs` config to use precompiled `ffmpeg` in order to include the plugin in Membrane SDK and make using it more straightforward
- Replace deprecated `channels` field with `AVChannelLayout` struct
- Prepare for releasing


Qizot (changes requested):
Please add PR description, what are you trying to achieve and why.

Noarkhh (changes requested):
Readme should be updated like other packages that now use precompiled dependencies, otherwise looks good :)

Qizot (changes requested):
I would expect a minor version bumped instead of the patch.

Qizot (approved):

Noarkhh (approved):

PR #84, opened 2023-12-29
Add onMetaData command support

Qizot:
This PR adds support for `onMetaData` AMF command which is present instead of `@setDataFrame` for certain clients e.g. nginx-rtmp module.

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/messages/on_meta_data.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/messages/on_meta_data.ex
@@ -1,0 +1,74 @@
+defmodule Membrane.RTMP.Messages.OnMetaData do
+  @moduledoc """
+  Defines the RTMP `onMetaData` command (sent by nginx client).
+  """
+
+  @behaviour Membrane.RTMP.Message
+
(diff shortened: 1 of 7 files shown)


lib/membrane_rtmp_plugin/rtmp/source/messages/on_meta_data.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  @behaviour Membrane.RTMP.Message
+
+  alias Membrane.RTMP.AMF.Encoder
+
+  defstruct ~w(
+    width height duration framerate video_data_rate video_codec_id audio_data_rate audio_codec_id
+  )a
+
+  @attributes_to_keys %{
```
balins, line 14: Nit: hardcoding keys-to-attributes results in a bit shorter notation. Also, you could `defstruct Map.keys(keys_to_attributes)`

balins (approved):

balins:
Also @Qizot let's update `bundlex` to the newest version before merging. The currently pinned one has a bug that breaks Elixir releases no matter if the user uses precompiled of pkg config deps.

PR #87, opened 2024-02-28
Handle socket closed by client before start of stream arrives in RTMP Source Bin

nickdichev-firework:
I had a flow in my application where the client was closing the socket before the start of stream was being sent. That was causing the RTMP source to send a message `:unexpected_socket_closed` to the parent, which in my case is the RTMP Source Bin. https://github.com/membraneframework/membrane_rtmp_plugin/blob/v0.22.1/lib/membrane_rtmp_plugin/rtmp/source/source.ex#L228

This message is not being handled by the bin so it causes a crash in my application. I have added handling for this case and decided to send the message to the bin's parent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The package can be installed by adding `membrane_rtmp_plugin` to your list of de
 ```elixir
 def deps do
   [
-	  {:membrane_rtmp_plugin, "~> 0.22.1"}
+	  {:membrane_rtmp_plugin, "~> 0.23.0"}
   ]
 end
 ```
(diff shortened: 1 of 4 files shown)


Qizot (commented):
Hey, thank you for the fix!

Could you please bump the version in `mix.exs` and in `README.md` (the patch version)?

Qizot (approved):

nickdichev-firework:
@Qizot sure, it is no problem, however, I'm not sure the patch version is the correct one to bump. Since now a new message is being sent to the parent pipeline if it is not explicitly handled it can crash the pipeline. What do you think?

Qizot:
@nickdichev-firework you're right. Bump the minor then :)

nickdichev-firework:
@Qizot I've sent a commit to bump the version.

PR #95, opened 2024-07-25
SourceBin dynamic pads

bartkrak:
solves https://github.com/membraneframework/membrane_core/issues/813

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
@@ -172,7 +172,9 @@ defmodule Membrane.RTMP.SourceBin do
       |> Enum.count(fn pad_ref -> Pad.name_by_ref(pad_ref) == name end)
 
     if count > 1 do
-      raise("Linking more than one #{name} output pad to #{inspect(__MODULE__)} is not allowed")
+      raise(
+        "Linking more than one #{inspect(name)} output pad to #{inspect(__MODULE__)} is not allowed"
+      )
     end
 
     :ok


lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
```diff
@@ -42,32 +42,100 @@ defmodule Membrane.RTMP.SourceBin do
...
+
+    if video_pad_ref != nil do
+      {[
+         spec: [
+           get_child(:demuxer)
+           |> via_out(pad_ref)
+           |> child(:video_parser, Membrane.H264.Parser)
+           |> get_child(:funnel_video)
+         ]
+       ], state}
```
FelonEkonom, line 133: If a spec is in form of `[element]`, it can be just `element`

lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
```diff
@@ -42,32 +42,100 @@ defmodule Membrane.RTMP.SourceBin do
...
+          |> via_out(state.demuxer_video_pad_ref)
+          |> child(:video_parser, Membrane.H264.Parser)
+          |> bin_output(pad)
+        ]
+      else
+        [
+          child(:funnel_video, Membrane.Funnel)
+          |> bin_output(pad)
+        ]
+      end
```
FelonEkonom, line 97: Raise, if there is more than 1 audio/video pad

FelonEkonom (changes requested):

varsill (changes requested):
General thought - I think it would be easier to understand if we would always spawn the `funnels` on bin's startup, and then link appropriate outputs when they ready. We could then move that "linking" to some function, like `link_audio` and `link_video`, so that not to repeat the code. 
Another benefit is that we would always have the same structure of the bin, no matter what is the order in which the callbacks were invoked, which definitely would simplify debugging (for instance, if there was some bug in the funnel)

lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
```diff
@@ -111,4 +134,47 @@ defmodule Membrane.RTMP.SourceBin do
...
+  end
+
+  defp assert_pad_count!(name, ctx) do
+    count =
+      ctx.pads
+      |> Map.keys()
+      |> Enum.count(fn pad_ref -> Pad.name_by_ref(pad_ref) == name end)
+
+    if count > 1 do
+      raise("Linking more than one #{name} output pad to #{__MODULE__} is not allowed")
```
varsill, line 175: [NIT] If you don't add `inspect()` it will print something like `Elixir.Membrane.RTMP.SourceBin` instead of `Membrane.RTMP.SourceBin`
```suggestion
      raise("Linking more than one #{name} output pad to #{inspect(__MODULE__)} is not allowed")
```

varsill (approved):
Looks much more clear now. Perhaps you could bump version to v0.25.0 and release it after it's merged?

lib/membrane_rtmp_plugin/rtmp/source/source_bin.ex
```diff
@@ -180,9 +135,46 @@ defmodule Membrane.RTMP.SourceBin do
...
+  end
+
+  defp assert_pad_count!(name, ctx) do
+    count =
+      ctx.pads
+      |> Map.keys()
+      |> Enum.count(fn pad_ref -> Pad.name_by_ref(pad_ref) == name end)
+
+    if count > 1 do
+      raise("Linking more than one #{name} output pad to #{inspect(__MODULE__)} is not allowed")
```
FelonEkonom, line 175: Wrap `name` into `inspect()`

FelonEkonom (approved):

PR #96, opened 2024-08-07
Refactor RTMP namespace

varsill:
This PR:
* bumps version to v0.24.0
* Changes namespace `RTMP.Server` into `RTMPServer`
* Move the `parse_url` utils function to RTMPServer
* Fixes a bug with `parse_url` so that it now support URLs without app and stream_key
* Unifies `RTMPServer.ClientHandler` and `RTMPServer.ClientHandlerBehaviour`
* Renames `Source.ClientHandler` into `Source.ClientHandlerForSource`

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/source/client_handler_for_source.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/source/client_handler_for_source.ex
@@ -1,4 +1,4 @@
-defmodule Membrane.RTMP.Source.ClientHandlerForSource do
+defmodule Membrane.RTMP.Source.ClientHandlerImpl do
   @moduledoc """
   An implementation of `Membrane.RTMPServer.ClienHandlerBehaviour` compatible with the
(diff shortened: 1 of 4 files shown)


lib/membrane_rtmp_plugin/rtmp/source/client_handler_for_source.ex
```diff
@@ -1,10 +1,10 @@
-defmodule Membrane.RTMP.Source.ClientHandler do
+defmodule Membrane.RTMP.Source.ClientHandlerForSource do
```
mat-hek, line 1: I understand it's to avoid confusion with the behaviour, but I'd leave it as is, just never alias it. Alternatively let's call it `ClientHandlerImpl`

varsill: I'm up to the second option ;)

mat-hek (approved):

PR #98, opened 2024-09-18
Simplify client handler

varsill:


mat-hek (approved):
Update the type in https://github.com/membraneframework/membrane_rtmp_plugin/blob/master/lib/membrane_rtmp_plugin/rtmp_server.ex#L24 and we can merge

mat-hek (approved):

PR #100, opened 2024-10-03
Draft: Improve socket connection dropped and delete stream events

dmorn:
This is here to close https://github.com/membraneframework/membrane_core/issues/792.

- behaviour remains the same, on socket connection closed end_of_stream is forwarded
- a new notification was added from the source component: :stream_deleted. This is delivered when the DeleteStream is received from RTMP (which signals the actual termination of a stream). Clients can decide how to react.

I'll keep it a draft waiting for your feedback (and some extensive tests on our side).

\cc @mat-hek

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp_server.ex
+++ b/lib/membrane_rtmp_plugin/rtmp_server.ex
@@ -27,8 +27,9 @@ defmodule Membrane.RTMPServer do
           port: :inet.port_number(),
           use_ssl?: boolean(),
           name: atom() | nil,
-          handle_new_client: (client_ref :: pid(), app :: String.t(), stream_key :: String.t() ->
-                                client_behaviour_spec()),
+          handle_new_client:
+            (client_ref :: pid(), app :: String.t(), stream_key :: String.t() ->
+               client_behaviour_spec()),
           client_timeout: Membrane.Time.t()
         ]
 
--- a/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
+++ b/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
@@ -40,7 +40,7 @@ defmodule Membrane.RTMPServer.ClientHandler do
 
   @doc """
   Callback invoked when the socket connection is terminated. In normal
-  conditions, delete_stream is called before this one. If delete_stream
+  conditions, `handle_delete_stream` is called before this one. If delete_stream
   is not called and connection_closed is, it might just mean that the
   connection was lost.
 
--- a/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
+++ b/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
@@ -34,12 +34,19 @@ defmodule Membrane.RTMPServer.ClientHandler do
   @callback handle_data_available(payload :: binary(), state :: state()) :: state()
 
   @doc """
-  The callback invoked when the client served by given client handler
-  stops sending data.
-  (for instance, when the remote client deletes the stream or
-  terminates the socket connection)
+  Callback invoked when the RMTP stream is finished.
   """
-  @callback handle_end_of_stream(state :: state()) :: state()
+  @callback handle_delete_stream(state :: state()) :: state()
+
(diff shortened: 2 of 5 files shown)


lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
```diff
@@ -34,12 +34,19 @@ defmodule Membrane.RTMPServer.ClientHandler do
...
-  (for instance, when the remote client deletes the stream or
-  terminates the socket connection)
+  Callback invoked when the RMTP stream is finished.
   """
-  @callback handle_end_of_stream(state :: state()) :: state()
+  @callback handle_delete_stream(state :: state()) :: state()
+
+  @doc """
+  Callback invoked when the socket connection is terminated. In normal
+  conditions, delete_stream is called before this one. If delete_stream
```
varsill, line 43:
```suggestion
  conditions, `handle_delete_stream` is called before this one. If delete_stream
```

lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
```diff
@@ -34,12 +34,19 @@ defmodule Membrane.RTMPServer.ClientHandler do
...
+  is not called and connection_closed is, it might just mean that the
+  connection was lost.
```
varsill, line 45:
```suggestion
  connection was lost (for instance because TCP socket is closed unexpectedly)
```

varsill (commented):
Hello!
Thanks for the contribution. It definitely makes sense to distinguish between "gracefull" shutdown with `DeleteStream` and some unexpected one.

I am just slightly concerned about the fact that `:stream_deleted` is not synchronized with an actual stream (perhaps an event would be better here?)
I also wonder how do you mean to use this `:stream_deleted` notification. Do you plan to set some flag upon receiving `:stream_deleted` notification in your parent element and than, depending on value of that flag, react differently when `handle_element_end_of_stream` is called?
If so, I think it is OK to leave the notification, but in a final run I would prefer to add some generic event to membrane core, that we could send along `:end_of_stream` to mark a "graceful" end of stream.

varsill:
>Our pipeline is pretty long and in case of an event we would have to pass it over the whole pipeline. A notification allows us to "jump" over to the concerned element. In our case, we turn the output HLS playlist into VOD only if we receive both the notification and end of stream.

I see, in this case it shouldn't be a problem at all. The problem could encounter if you relied just on the notification (there wouldn't be any guarantee that the stream in the rest of the pipeline is already processed when you turn the playlist into VOD).
Concerning event passing, I don't think it would be a problem to pass it over the whole pipeline, as default implementation of the `handle_event` callback in elements is forwarding incoming events further. But for now I think it's OK just to send the notification and use it along `:end_of_stream` ;)

dmorn:
@varsill I did not know that! If that's the case, an event might be even better, let me know what you're more keen on merging. About the linter, I'm running `mix format` in the proj but it's still complaining, am I missing something? Side note: can we move a bit faster in the merge process? I have other 2 PR's coming and it would be great if we could do all of this today :D

varsill:
Hi @dmorn , sure, sorry for late reply, I am OK with merging what we currently have (the notification sent from the source). Concerning the formatter, I think there might be a mismatch in the version of elixir/mix you are using (could you tell me what do you use?)

lib/membrane_rtmp_plugin/rtmp_server.ex
```diff
@@ -27,8 +27,9 @@ defmodule Membrane.RTMPServer do
           port: :inet.port_number(),
           use_ssl?: boolean(),
           name: atom() | nil,
-          handle_new_client: (client_ref :: pid(), app :: String.t(), stream_key :: String.t() ->
-                                client_behaviour_spec()),
+          handle_new_client:
+            (client_ref :: pid(), app :: String.t(), stream_key :: String.t() ->
+               client_behaviour_spec()),
```
varsill, line 32:
```suggestion
          handle_new_client:
            (client_ref :: pid(), app :: String.t(), stream_key :: String.t() ->
                                  client_behaviour_spec()),
```

PR #102, opened 2024-10-28
Send RTMP messages to receiver_pid

lastcanal:
Hello! Upon upgrading to v0.26 I've lost access to the RTMP messages which were previously available to the Validator. 
 I saw that `Membrane.RTMP.MessageHandler` already has an unused `receiver_pid` in it's state, so I used that for sending messages to the receiver. 

To set the `receiver_pid` users can return a three-tuple from the `handle_new_client` callback like this:

```elixir
{Membrane.RTMP.Source.ClientHandlerImpl, %{}, receiver_pid}
```

RTMP messages that get sent are simply forwareded `Membrane.RTMP.Messages` structs, not wrapped in any 'scoping' tuple. The `OnDataFrame` message is particularly important to me, as I am using it to calculate frame dimensions for transcoding with `membrane_ffmpeg_swscale_plugin`.

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp/message_handler.ex
+++ b/lib/membrane_rtmp_plugin/rtmp/message_handler.ex
@@ -161,20 +161,23 @@ defmodule Membrane.RTMP.MessageHandler do
   end
 
   # A message containing stream metadata
-  defp do_handle_client_message(%Messages.SetDataFrame{} = _data_frame, _header, state) do
+  defp do_handle_client_message(%Messages.SetDataFrame{} = set_data_frame, _header, state) do
+    if pid = state.receiver_pid, do: send(pid, set_data_frame)
     {:cont, state}
   end
 
-  defp do_handle_client_message(%Messages.OnMetaData{} = _on_meta_data, _header, state) do
+  defp do_handle_client_message(%Messages.OnMetaData{} = on_meta_data, _header, state) do
+    if pid = state.receiver_pid, do: send(pid, on_meta_data)
     {:cont, state}
   end
 
   # According to ffmpeg's documentation, this command should prepare the server to receive media streams
   # We are simply acknowledging the message
-  defp do_handle_client_message(%Messages.FCPublish{}, _header, state) do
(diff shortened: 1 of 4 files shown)


varsill:
Hello @lastcanal ! 
That's a good observation, that it used to be possible to do it that way, but it'a not possible anymore.
But could you provide some more context why do you need that "OnDataFrame" message (BTW - didn't you mean `OnMetaData` message? I think all the necessary information (like resolution etc.) should be available in the stream format.
Another thing is that I guess adding a new callback in the `ClientHandlerBehaviour` (like `handle_rtmp_message`) which would be called with each message received after RTMP handshake for a given client, aligns to the current architecture of the RTMP server in greater degree. 
On the other hand, it would mean that for your case you would need to implement a custom behaviour.
WDYT about such an approach?

lastcanal:
Hello! Sorry, I meant `SetDataFrame`, although `OnMetaData` should work too: I really only need width, height, and framerate for scaling video tracks while keeping the aspect ratio intact.

I took your suggestion and tried to use the stream format, but unfortunately `framerate` is always set to `nil`... Perhaps this is a bug? Here is the stream format for my OBS RTMP stream intercepted by a `Tee.Parallel`:

```elixir
%Membrane.H264{
  width: 1280,
  height: 720,
  profile: :high,
  alignment: :au,
  nalu_in_metadata?: true,
  framerate: nil,
  stream_structure: {:avc3, <<1, 100, 0, 32, 255, 224, 0>>}
}
```

Sending messages through the client handler is a great idea and i managed to get that working. I've pushed a new branch #103 with the new `handle_rtmp_message` callback. Thanks!

varsill:
@lastcanal - sure, I understand. Indeed, the framerate is missing, because the `H264` stream format is created by the FLV Demuxer (a part of the SourceBin) and FLV demuxer is apparently unable to read that information from FLV, and at the same time metadata sent within RTMP messages is lost.
There are is a couple of possible approaches to that problem, I've described them here: https://github.com/membraneframework/membrane_core/issues/903

PR #103, opened 2024-10-29
Send RTMP messages to client handler

lastcanal:
Adds optional callback `handle_rtmp_message/2` to ClientHandlerBehaviour

I've put together a demo (https://github.com/membraneframework/membrane_demo/pull/287) that uses a [custom client handler](https://github.com/lastcanal/membrane_demo/blob/20fb447764490d0db2ed86ec284d45cac8323304/rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/client_handler.ex#L57-L60) with the proposed `handle_rtmp_message` callback to forward RTMP messages to a pipeline that creates [multi-variant HLS playlists](https://github.com/lastcanal/membrane_demo/blob/20fb447764490d0db2ed86ec284d45cac8323304/rtmp_to_adaptive_hls/lib/rtmp_to_adaptive_hls/pipeline.ex#L42-L81).

### DIFF
--- a/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
+++ b/lib/membrane_rtmp_plugin/rtmp_server/client_handler.ex
@@ -193,8 +193,7 @@ defmodule Membrane.RTMPServer.ClientHandler do
     state = Enum.reduce(events, state, &handle_event/2)
 
     state =
-      if state.handler &&
-           state.handler_state &&
+      if state.notified_about_client? &&
            Kernel.function_exported?(state.handler, :handle_rtmp_message, 2) do
         new_handler_state =
           Enum.reduce(messages, state.handler_state, fn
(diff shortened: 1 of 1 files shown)


varsill:
Hello @lastcanal !
Thank you for your contribution.
After some consideration I've come to the conclusion that calling `handle_rtmp_message` only for the messages after the handshake is finished might be slightly inconsistent (at the same time, we cannot do it before the handshake finishes as we don't have client handler then).
I guess it would be better to add `handle_metadata` callback and have an explicit list of messages for which this callback would be called.
I've opened a PR: https://github.com/lastcanal/membrane_rtmp_plugin/pull/1 with the changes I suggested.

lastcanal:
Thank you @varsill! I've merged your changes into my branch. Your way makes the most sense, especially considering there is no need for the audio and video messages bogging down the client handler. I appreciate your help!

varsill:
No problem, thank you very much @lastcanal !

varsill (approved):

PR #107, opened 2024-10-31
Handle stream_deleted child notification in SourceBin

lastcanal:
Hello! I am getting a `FunctionClauseError` in `RTMP.SourceBin` when an RTMP stream ends before the pipeline is terminated. It looks like `RTMP.SourceBin` is not setup to handle the `:stream_deleted` child notification.

This PR:
  - Updates the 'happy path' test to wait for the ffmpeg stream to end before terminating the test pipeline.
  - Handles the `stream_deleted` child notification from `Source` to `SourceBin` by forwarding to it's parent.


lastcanal:
Looks like this was already handled in #106 but I didn't see it before creating this PR :). I've rebased my changes to only include the failing test case.

PR #113, opened 2025-06-16
Add RTMPS support

olegokunevych:


lib/membrane_rtmp_plugin/rtmp_server/config.ex
```diff
@@ -0,0 +1,363 @@
...
+
+  @doc """
+  Validates that required SSL files exist and options are valid.
+  Set validate_files to false to skip file existence checks (useful for testing).
+  """
+  @spec validate_ssl_options(ssl_options(), validate_files :: boolean()) :: ssl_options()
+  def validate_ssl_options(opts, validate_files \\ true) do
+    opts = validate_certificate_files(opts, validate_files)
+    opts = validate_ssl_configuration(opts)
+    opts
```
varsill, line 146:
```suggestion
    validate_certificate_files(opts, validate_files)
    |> validate_ssl_configuration(opts)
```

lib/membrane_rtmp_plugin/rtmp_server/config.ex
```diff
@@ -0,0 +1,363 @@
...
+        Keyword.put(opts, key, priv_path)
+      else
+        # Keep original path
+        opts
+      end
+    end
+  end
+
+  @spec validate_certificate_files(ssl_options(), boolean()) :: ssl_options()
+  defp validate_certificate_files(opts, validate_files) do
```
varsill, line 264: Couldn't we just pattern match on` validate_files` being `true` here and directly call `raise ArgumentError` (instead of `validate_singe_cert_file` etc.) in case some file is missing?

lib/membrane_rtmp_plugin/rtmp_server/config.ex
```diff
@@ -0,0 +1,363 @@
...
+
+    # Validate log level
+    if log_level = opts[:log_level] do
+      valid_levels = [:none, :error, :warning, :notice, :info, :debug, :all]
+
+      unless log_level in valid_levels do
+        raise ArgumentError,
+              "Invalid SSL log level: #{inspect(log_level)}. " <>
+                "Valid levels are: #{inspect(valid_levels)}"
+      end
```
varsill, line 358: do we really need to validate it here? I would leave it up to the `:ssl` Erlang module

lib/membrane_rtmp_plugin/rtmp_server/listener.ex
```diff
@@ -5,47 +5,120 @@ defmodule Membrane.RTMPServer.Listener do
...
+        end
+
+        # Additional validation for certificate files
+        if ssl_opts[:certfile] && !File.exists?(ssl_opts[:certfile]) do
+          raise ArgumentError, "SSL certificate file does not exist: #{ssl_opts[:certfile]}"
+        end
+
+        if ssl_opts[:keyfile] && !File.exists?(ssl_opts[:keyfile]) do
+          raise ArgumentError, "SSL key file does not exist: #{ssl_opts[:keyfile]}"
+        end
```
varsill, line 58: We already check that `ssl_opts is non `nil` (and otherwise we `raise`) so I don't think we need that check here as well

lib/membrane_rtmp_plugin/rtmp_server/listener.ex
```diff
@@ -5,47 +5,120 @@ defmodule Membrane.RTMPServer.Listener do
...
-          active: false,
-          certfile: certfile,
-          keyfile: keyfile
-        ]
+        ssl_opts = Config.get_ssl_listen_options(Map.get(options, :ssl_options, []), true)
+
+        Logger.debug("SSL options for listen: #{inspect(ssl_opts)}")
+
+        # SSL listen requires at least certfile and keyfile
+        unless ssl_opts[:certfile] && ssl_opts[:keyfile] do
```
varsill, line 29: Aren't these checks already made in `Config.get_ssl_listen_options`?

lib/membrane_rtmp_plugin/rtmp_server/config.ex
```diff
@@ -0,0 +1,363 @@
...
+        :password,
+        :versions
+      ])
+
+    # Ensure we have the minimum required options
+    if ssl_listen_opts == [] do
+      []
+    else
+      ssl_listen_opts
+    end
```
varsill, line 100: I think we can remove it and return just `ssl_listen_opts` ;)

varsill (changes requested):

olegokunevych:
@varsill thank you for code review, I've found out it useful and committed corresponding changes. Let me know if we need to add/modify something else in this PR.
PS I am wondering that CI format checker is failing, I've done `mix format` locally

olegokunevych:
@mat-hek @varsill I've simplified an API and removed all the validations for SSL options. Let me know what you think about the current approach.
IMHO it should be responsibility of library to validate passed options, but we can start with this implementation

mat-hek (approved):

varsill:
Once again, thank you very much @olegokunevych for your contribution! 🥇

PR #118, opened 2025-10-21
Implement spec-compliant chunk-stream-id serialization/deserialization

dmorn:
The RTMP server was crashing when receiving streams from certain hardware encoders with the following error:

```
[warning] Unexpected header type when combining body chunks: 1
[error] GenServer #PID<0.1014.0> terminating
** (KeyError) key :timestamp not found in: nil
    (membrane_rtmp_plugin 0.29.1) lib/membrane_rtmp_plugin/rtmp/header.ex:114: Membrane.RTMP.Header.deserialize/2
    (membrane_rtmp_plugin 0.29.1) lib/membrane_rtmp_plugin/rtmp/message_parser.ex:180: Membrane.RTMP.MessageParser.read_frame/3
```
The crash was caused by **incomplete support for RTMP chunk interleaving**. Further, only chunk-stream-id singe byte serialization/deserialization was implemented.

This PR adds support for 1/2/3 byte type stream chunk identifiers and handles the situation where chunks from different streams are interleaved together. Fully backward compatible.


lib/membrane_rtmp_plugin/rtmp/source/source.ex
```diff
@@ -176,18 +176,31 @@ defmodule Membrane.RTMP.Source do
...
 
+  @impl true
+  def handle_info(
+        {:client_ref, _client_ref, app, stream_key},
+        _ctx,
+        %{mode: :builtin_server} = state
+      )
+      when app == state.app and stream_key == state.stream_key do
+    # Duplicate connection on same app/stream_key - ignore it
+    Logger.warning("Duplicate client connection detected on /#{app}/#{stream_key}, ignoring")
```
varsill: Perhaps we could log the `client_ref` here so that it would be possible to distinguish which one gets discarded?

varsill (approved):
Thanks for the PR!
I left one minor comment.
I also wonder if it would be possible for you to add a single integration test checking the interleaved chunks behaviour (especially if you already have some stream dump we could use as a fixture).

dmorn:
Thanks! I addressed the logging comment by including both the discarded `client_ref` and the already-handled one in the warning.

I also added an integration-style parser regression test based on a real Blackmagic Web Presenter 4K capture. The fixture stores post-handshake client→server RTMP TCP reads with original read boundaries preserved, which turned out to be important for reproducing the issue. I kept it constrained (~800 KB) instead of committing a full 4K media recording.

The interesting bit is that replaying this fixture leaves leftover buffered data on `master`, while this branch fully consumes it, so it gives us a real-world regression check for the parser behavior we’re fixing.

varsill:
Thanks for contribution @dmorn ! I will release it together with https://github.com/membraneframework/membrane_rtmp_plugin/pull/122, is that OK?

dmorn:
Of course! I'll prepare in next week (tuesday) @varsill!

PR #122, opened 2026-03-19
Fix/client handler leak

dmorn:
If a client does TCP health-checks on a listener based on this library, we leak ClientHandler processes at each call. This addresses that by closing the hander. In is built on top of #118


varsill:
As far as I can see the actual change in this PR is stopping the client handler GenServer's, let's rebase the PR after https://github.com/membraneframework/membrane_rtmp_plugin/pull/118 is merged and it will be fine to merge this one too 👍

